PR #42, opened 2021-10-15
Navigation component + Page/Layout concept

oshi97:
This PR adds the navigation component, and also reworks the react-router usage into a Page and Layout system.

For the Navigation resizing, I tried to use resize observer, first with https://www.npmjs.com/package/@react-hook/resize-observer and then with ResizeObserver directly. The third party hook did not seem to work well for our use case,
and for using ResizeObserver directly, it turns out the TypeScript types for ResizeObserver are not included by default,
and I felt a little apprehensive to include third party types for it, though it would probably be fine.
https://github.com/Microsoft/TypeScript/issues/28502

J=SLAP-1558
TEST=manual

test resizing the page and seeing tabs appear and disappear
test that switching tabs will run searches


sample-app/src/pages/VerticalSearchPage.tsx
```diff
@@ -0,0 +1,107 @@
...
+  c_employeeDepartment: {
+    label: 'Employee Department!'
+  }
+}
+
+export default function VerticalSearchPage(props: {
+  verticalKey: string
+}) {
+  const answersActions = useAnswersActions();
+  useLayoutEffect(() => {
```
yen-tt, line 57: I believe useLayoutEffect is fired before DOM mutation? I think we don't need to wrap this around useLayoutEffect. I think something like this would work:
```export default function VerticalSearchPage(props: {
  verticalKey: string
}) {
  answersActions.setVerticalKey(props.verticalKey);
  answersActions.executeVerticalQuery();
  ...```

File: sample-app/src/pages/VerticalSearchPage.tsx [hunk 2]
```diff
@@ -0,0 +1,107 @@
...
+    answersActions.setState({
```
yen-tt, line 58: do we need to reset universal state here?

sample-app/src/pages/UniversalSearchPage.tsx
```diff
@@ -0,0 +1,29 @@
+import UniversalResults, { VerticalConfig } from '../components/UniversalResults';
+import { useLayoutEffect } from 'react';
+import { useAnswersActions } from '@yext/answers-headless-react';
+import '../sass/UniversalSearchPage.scss';
+
+const universalResultsFilterConfig = {
+  show: true
+};
+
+export default function UniversalSearchPage(props: { universalResultsConfig: Record<string, VerticalConfig>}) {
```
yen-tt, line 10: nit, I think maybe we should make an interface for the props, and have it include universalResultsFilterConfig in it as well, seems a little odd that one is passed in and the other is hardcoded in this file?

sample-app/src/components/Navigation.tsx
```diff
@@ -0,0 +1,99 @@
...
+          <div className='Navigation__menuLinks'>
+            {menuOpen && overflowLinks.map(l => renderLink(l, search))}
+          </div>
+        </div>
+      }
+    </nav>
+  )
+}
+
+function renderLink(linkData: LinkData, search: string) {
```
yen-tt, line 86: ooh could you update StandardSection.tsx to use this instead for 'View All' link to the vertical page?

oshi97: I took a look but the links are doing different things. This one is preserving query params, while the one in StandardSection is adding a new one. I thought about a wrapper component for both, but it would end up being just a pass through to NavLink. StandardSection also probably doesn't need a NavLink, and a regular Link might be better.

sample-app/src/BaseRouter.tsx
```diff
@@ -0,0 +1,38 @@
...
+
+interface RouteData {
+  path: string
+  page: JSX.Element
+  exact?: boolean
+}
+
+interface PageProps {
+  Layout?: ComponentType<{ page: JSX.Element }>
+  routes: (RouteData | null)[]
```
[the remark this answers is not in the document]
oshi97: actually I'll just remove it, it's not like it's some impossibly hard to see idea

sample-app/src/BaseRouter.tsx
```diff
@@ -0,0 +1,35 @@
...
+import { BrowserRouter as Router, Switch, Route } from 'react-router-dom';
+
+interface RouteData {
+  path: string
+  page: JSX.Element
+  exact?: boolean
+}
+
+interface PageProps {
+  Layout?: ComponentType<{ page: JSX.Element }>
```
cea2aj, line 11: do we have a convention behind using camelCase vs PascalCase for props?

cea2aj: makes sense!

sample-app/src/pages/VerticalSearchPage.tsx
yen-tt: hm ok, I see the flickering. that's fine then!

sample-app/src/Layout.tsx
```diff
@@ -0,0 +1,29 @@
...
+    to: '/',
+    label: 'All'
+  },
+  ...Object.entries(universalResultsConfig).map(([verticalKey, config]) => ({
+    to: verticalKey,
+    label: config.label || verticalKey
+  }))
+]
+
+export default function Layout({ page }: { page: JSX.Element }) {
```
cea2aj, line 17: can you add some jsdocs to help describe the responsibility of these different components?

sample-app/src/BaseRouter.tsx
```diff
@@ -0,0 +1,35 @@
...
+  routes: RouteData[]
+}
+
+export default function BaseRouter({ Layout, routes }: PageProps) {
+  const pages = routes.map(routeData => {
+    const { path, page, exact } = routeData;
+    if (Layout) {
```
cea2aj, line 18: I'm wondering if the Layout props adds much value to the router. I would expect that a router would simply render what you give it at a given route. I briefly looked into the react router and I didn't see any sort of layout concept, but it's possible I missed something there. What are some of the reasons that lead you to adding Layout to the router?

sample-app/src/BaseRouter.tsx
```diff
@@ -0,0 +1,35 @@
...
+interface RouteData {
+  path: string
+  page: JSX.Element
+  exact?: boolean
+}
+
+interface PageProps {
+  Layout?: ComponentType<{ page: JSX.Element }>
+  routes: RouteData[]
+}
```
cea2aj, line 13: Would it make sense to have a Layout interface defined in typescript? Then the Layout.tsc could implement that and we'd maybe want to rename that Layout since it is a concrete implemenation of the interface?

sample-app/src/BaseRouter.tsx
oshi97: I wanted to abstract away react-router details for the the kind of use case we would expect theme 2.0 to have, and wanted a way for a given route's page to have shared components. We had some idea of "layouts" for jambo that eventually fizzled out, but I thought made sense here

sample-app/src/BaseRouter.tsx
oshi97: added!

sample-app/src/Layout.tsx
oshi97: added some and also renamed BaseRouter -> PageRouter

sample-app/src/pages/VerticalSearchPage.tsx
```diff
@@ -0,0 +1,107 @@
...
+      ...answersActions.state,
+      universal: {}
```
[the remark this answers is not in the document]
oshi97: maybe we can wait until we get to the url param/doing stuff on load PRs

sample-app/src/Layout.tsx
```diff
@@ -0,0 +1,34 @@
...
+  ...Object.entries(universalResultsConfig).map(([verticalKey, config]) => ({
+    to: verticalKey,
+    label: config.label || verticalKey
+  }))
+]
+
+/**
+ * A LayoutComponent that provides a SearchBar and Navigation tabs to a given page.
+ */
+const Layout: LayoutComponent = ({ page }) => {
```
cea2aj, line 21: super nit: It might be easy to mix up the `Layout` here and the `Layout` in PageProps because one is a concrete implementation and the other simply obeys the `LayoutComponent` interface. What do you think about naming this `StandardLayout` in order to help differentiate them?

sample-app/src/components/Navigation.tsx
```diff
@@ -0,0 +1,99 @@
...
+  const { to, label } = linkData;
+  return (
+    <NavLink
+      key={to}
+      className='Navigation__link'
+      activeClassName='Navigation__link--currentRoute'
+      to={`${to}${search}`}
```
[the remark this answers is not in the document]
oshi97: so it'll look something like `?query=tom`

sample-app/src/Layout.tsx
oshi97: yeah I like that! I also moved it to the `src/pages` folder

sample-app/src/components/Navigation.tsx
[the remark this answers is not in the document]
oshi97: I renamed search to queryParams inside renderLink!

PR #49, opened 2021-10-28
Add commonjs build

cea2aj:
Add a commonjs build so that the library is compatible with node

There is a warning about useLayoutEffect since that is not compatible with server-side-rendering, however it still works. It will lead to a mismatch between the initial non-hydrated UI and the intended UI. If we do want to fully support that use case, there is more info here: https://gist.github.com/gaearon/e7d97cdf38a2907924ea12e4ebdf3c85, however that is out of scope of this item

J=SLAP-1666
TEST=manual

Use create-next-app and add answers-headless-react to it and test that a simple universal search works.

### DIFF
--- a/THIRD-PARTY-NOTICES
+++ b/THIRD-PARTY-NOTICES
@@ -95,7 +95,7 @@ MIT License
 The following NPM packages may be included in this product:
 
  - @yext/answers-core@1.3.2
- - @yext/answers-headless-react@0.4.0-beta.0
+ - @yext/answers-headless-react@0.4.0-beta.1
  - @yext/answers-headless@0.1.0-beta.0
 
 These packages each contain the following license and notice below:
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@yext/answers-headless-react",
-  "version": "0.4.0-beta.0",
+  "version": "0.4.0-beta.1",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
       "name": "@yext/answers-headless-react",
-      "version": "0.4.0-beta.0",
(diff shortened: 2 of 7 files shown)


oshi97:
This is what react-redux does for react on the server side
https://github.com/reduxjs/react-redux/blob/master/src/utils/useIsomorphicLayoutEffect.ts

tmeyer2115:
> This is what react-redux does for react on the server side, not saying we should do this just adding more context (it's just using useEffect in node environments like Connor also mentioned, the source code is almost identical to react-layout-effect) https://github.com/reduxjs/react-redux/blob/master/src/utils/useIsomorphicLayoutEffect.ts

@cea2aj, would the isomorphic `useLayoutEffect` or `useIsomorphicLayoutEffect` allow the static HTML to be generated properly and then the dynamic `useAnswersState` to function once the page is hydrated? I will check with Product about how much time we should be spending on SSR of React Components. I definitely see the value in having `answers-headless` work with Node, but am not fully sold on SSR of React.

cea2aj:
> @cea2aj, would the isomorphic `useLayoutEffect` or `useIsomorphicLayoutEffect` allow the static HTML to be generated properly and then the dynamic `useAnswersState` to function once the page is hydrated? I will check with Product about how much time we should be spending on SSR of React Components. I definitely see the value in having `answers-headless` work with Node, but am not fully sold on SSR of React.

No, those functions would just fix the warning in the console, but they wouldn't improve the capabilities of the initial render. We'd have to test out the components with SSR to see if the components look broken. At worst, we'll have to disable SSR of the components and let client side React render the components

tmeyer2115 (approved):

oshi97 (approved):

PR #50, opened 2021-11-02
Refactor StaticFilters

yen-tt:
Refactor StaticFilters and update AppliedFilters to follow the new public interface from answers-headless. 

- update StaticFilters to be a functional component
- update StaticFilters to use new interface from answers-headless changes in [this PR](https://github.com/yext/answers-headless/pull/42) instead of managing its own filter state.
- update filter utils to reflect the new interfaces from answers-headless
- update AppliedFilters to use answers-headless `setFilterOption` function instead of performing a click event

J=SLAP-1661
TEST=manual

- hook to local answers-headless changes, and start sample-app. See that filters work accordingly.

### DIFF
--- a/sample-app/src/components/AppliedFilters.tsx
+++ b/sample-app/src/components/AppliedFilters.tsx
@@ -79,7 +79,7 @@ function RemovableFilter({ filter }: {filter: DisplayableFilter }): JSX.Element 
       console.error(`Undefined filter collection id. Unable to remove this filter in AppliedFilters component:\n${JSON.stringify(filter)}`);
       return;
     }
-    answersAction.toggleFilterOption(filter.filterCollectionId, filter.filter, false);
+    answersAction.toggleFilterOption({ filter: filter.filter, selected: false }, filter.filterCollectionId);
     answersAction.executeVerticalQuery();
   }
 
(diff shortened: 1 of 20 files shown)


sample-app/src/components/AppliedFilters.tsx
```diff
@@ -75,7 +75,12 @@ function RemovableFilter({ filter }: {filter: DisplayableFilter }): JSX.Element
   }
 
   const onRemoveStaticFilterOption = () => {
-    document.getElementById(`${filter.filter.fieldId + "_" + filter.filter.value}`)?.click();
+    if (!filter.filterCollectionId) {
```
oshi97, line 78: do we need this console.error if answers-headless also logs an error for this?

yen-tt: filterCollectionId can't be required for DisplayableFilter since DisplayableFilter can also be of type Facet or NLP filters, which doesn't have filterCollectionId. Typescript would yell that `filter.filterCollectionId` can be undefined, which is not allow for function setFilterOption. So a check would be needed here anyway. I think it's ok to keep? Or I can remove the console.error line to avoid duplicate message

oshi97 (approved):

PR #51, opened 2021-11-02
Add labels to applied static filters

nmanu1:
Allow users to specify group labels for applied static filters. `DecoratedAppliedFilters` accepts a mapping of `fieldId` to group label. If no label is provided, it defaults back to using the `fieldId` as the group label.

J=SLAP-1658
TEST=manual

Check that specifying a group label for a static filter fieldId is reflected when a filter with that fieldId is applied

### DIFF
--- a/sample-app/src/components/DecoratedAppliedFilters.tsx
+++ b/sample-app/src/components/DecoratedAppliedFilters.tsx
@@ -3,14 +3,16 @@ import { AppliedQueryFilter } from "@yext/answers-core";
 import { useAnswersState } from '@yext/answers-headless-react';
 import { GroupedFilters } from '../models/groupedFilters';
 import { getGroupedAppliedFilters } from '../utils/appliedfilterutils';
-import { StaticFiltersLabelConfig } from "./StaticFilters";
 
 export interface DecoratedAppliedFiltersConfig {
   showFieldNames?: boolean,
   hiddenFields?: Array<string>,
   labelText?: string,
   delimiter?: string,
-  staticFiltersGroupLabels?: Record<string, StaticFiltersLabelConfig>,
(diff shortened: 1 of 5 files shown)


sample-app/src/components/DecoratedAppliedFilters.tsx
```diff
@@ -3,23 +3,25 @@ import { AppliedQueryFilter } from "@yext/answers-core";
...
 import { GroupedFilters } from '../models/groupedFilters';
 import { getGroupedAppliedFilters } from '../utils/appliedfilterutils';
+import { StaticFiltersLabelConfig } from "./StaticFilters";
 
 export interface DecoratedAppliedFiltersConfig {
   showFieldNames?: boolean,
   hiddenFields?: Array<string>,
   labelText?: string,
   delimiter?: string,
+  staticFiltersGroupLabels?: Record<string, StaticFiltersLabelConfig>,
```
oshi97, line 13: Would it be simpler if this was just a mapping of fieldId to label? instead of a mapping of fieldId to an object with a single label property

nmanu1: Yes, that would be simpler. I did it this way to make it more clear what the mapping was between, but I could also just do that with documentation instead

oshi97: Ah ok, let's just go with the string to string mapping for now, that way people don't have to understand what a StaticFiltersLabelConfig is, if needed we can tweak it later

nmanu1: Updated!

oshi97 (approved):

PR #54, opened 2021-11-09
Update React Sample App according to new Headless State

yen-tt:
Update React sample app and tests based on the updated headless state

- update components to use the new vertical and universal state interfaces, and the new `SearchStatusState` and `QueryState`
- update `SearchBar` to manage its own autocomplete results in component state. Added a useAutocomplete hook
  - Note: autocomplete dropdown will flicker between old and recently trigger search results. This will be address through supporting async and non-async in `inputDropdown` in another item
- update `AppliedFilter` and `StaticFilter` to no longer use ID with headless public interface
- as discussed with product, any update to static Filter should reset facet to avoid invalid facet send as part of request. Added call to `resetFacets` on `handleFilterOptionChange`

J=SLAP-1693
TEST=manual

jest tests passed
smoke tested sample-app with new headless version
- see that universal search executed properly
- see that vertical search work with multiple static filters
- see that UI for loading state behave properly with universal/vertical search
- see that autocomplete display options properly

### DIFF
--- a/sample-app/src/components/InputDropdown.tsx
+++ b/sample-app/src/components/InputDropdown.tsx
@@ -12,7 +12,7 @@ interface Props {
   optionIdPrefix: string,
   onSubmit?: (value: string) => void,
   updateInputValue: (value: string) => void,
-  updateDropdown: () => Promise<void>,
+  updateDropdown: () => void,
   renderButtons?: () => JSX.Element | null,
   cssClasses: {
     optionContainer: string,
@@ -126,16 +126,16 @@ export default function InputDropdown({
         <input
           className={cssClasses.inputElement}
           placeholder={placeholder}
-          onChange={async evt => {
+          onChange={evt => {
             const value = evt.target.value;
             dispatch({ type: 'ShowOptions' });
             setLatestUserInput(value);
             updateInputValue(value);
-            await updateDropdown();
+            updateDropdown();
             setScreenReaderKey(screenReaderKey + 1);
           }}
-          onClick={async () => {
-            await updateDropdown();
+          onClick={() => {
+            updateDropdown();
             dispatch({ type: 'ShowOptions' });
             if (options.length || inputValue) {
               setScreenReaderKey(screenReaderKey + 1);
--- a/sample-app/src/components/InputDropdown.tsx
+++ b/sample-app/src/components/InputDropdown.tsx
@@ -12,7 +12,7 @@ interface Props {
   optionIdPrefix: string,
   onSubmit?: (value: string) => void,
   updateInputValue: (value: string) => void,
-  updateDropdown: () => void,
+  updateDropdown: () => Promise<void>,
   renderButtons?: () => JSX.Element | null,
   cssClasses: {
     optionContainer: string,
@@ -126,17 +126,17 @@ export default function InputDropdown({
         <input
           className={cssClasses.inputElement}
           placeholder={placeholder}
-          onChange={evt => {
+          onChange={async evt => {
             const value = evt.target.value;
             dispatch({ type: 'ShowOptions' });
             setLatestUserInput(value);
             updateInputValue(value);
-            updateDropdown();
+            await updateDropdown();
             setScreenReaderKey(screenReaderKey + 1);
           }}
-          onClick={() => {
+          onClick={async () => {
+            await updateDropdown();
             dispatch({ type: 'ShowOptions' });
-            updateDropdown();
             if (options.length || inputValue) {
               setScreenReaderKey(screenReaderKey + 1);
             }
--- a/sample-app/src/components/SearchBar.tsx
+++ b/sample-app/src/components/SearchBar.tsx
@@ -1,10 +1,11 @@
-import { useAnswersActions, useAnswersState, StateSelector, AutocompleteResult } from '@yext/answers-headless-react';
+import { useAnswersActions, useAnswersState, AutocompleteResult } from '@yext/answers-headless-react';
 import InputDropdown from './InputDropdown';
 import renderWithHighlighting from './utils/renderWithHighlighting';
 import { ReactComponent as MagnifyingGlassIcon } from '../icons/magnifying_glass.svg';
 import '../sass/SearchBar.scss';
 import '../sass/Autocomplete.scss';
 import LoadingIndicator from './LoadingIndicator';
+import { useRef, useState } from 'react';
 
 const SCREENREADER_INSTRUCTIONS = 'When autocomplete results are available, use up and down arrows to review and enter to select.'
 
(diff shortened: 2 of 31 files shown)


sample-app/src/components/SearchBar.tsx
```diff
@@ -19,17 +20,20 @@ interface Props {
...
-    ? state => state.vertical.autoComplete?.results
-    : state => state.universal.autoComplete?.results;
-  const autocompleteResults = useAnswersState(mapStateToAutocompleteResults) || [];
-  const isLoading = useAnswersState(state => state.vertical.searchLoading || state.universal.searchLoading);
-
-  function executeAutocomplete () {
+  const query = useAnswersState(state => state.query.input);
+  const [ autocompleteResults, setAutoCompleteResults ] = useState<AutocompleteResult[]>([]);
+  const isLoading = useAnswersState(state => state.searchStatus.isLoading);
+  const networkIds = useRef({ request: 0, response: 0 });
```
cea2aj, line 26: what do you think about `autocompleteNetworkIds`? We could also update `request` to `latestRequest`

cea2aj: I'm also wondering if we should update `response` to something like `responseInState` or `responseInUse`

sample-app/src/components/SearchBar.tsx
```diff
@@ -19,17 +20,20 @@ interface Props {
...
+    const requestId = ++networkIds.current.request;
+    let response = null;
     isVertical
-      ? answersActions.executeVerticalAutoComplete()
-      : answersActions.executeUniversalAutoComplete()
+      ? response = await answersActions.executeVerticalAutocomplete()
+      : response = await answersActions.executeUniversalAutocomplete();
+    if (requestId >= networkIds.current.response) {
+      setAutoCompleteResults(response?.results || []);
+      networkIds.current.response++;
```
cea2aj, line 35: Do we always increment this? Or should we set `networkIds.current.response` to the `requestid`?

sample-app/src/components/InputDropdown.tsx
```diff
@@ -12,7 +12,7 @@ interface Props {
   optionIdPrefix: string,
   onSubmit?: (value: string) => void,
   updateInputValue: (value: string) => void,
-  updateDropdown: () => void,
+  updateDropdown: () => Promise<void>,
```
cea2aj, line 15: Was there a problem that required making this async?

sample-app/src/components/SearchBar.tsx
yen-tt: ah yeah, should be to requestId!

sample-app/src/components/SearchBar.tsx
yen-tt: sounds good

sample-app/src/components/InputDropdown.tsx
yen-tt: yeah, since `onClick` calls updateDropdown and dispatch for 'showOptions', without async wait for updateDropdown, autocomplete would flicker between previous result shown and the latest result during render (i.e. showOptions dispatch will trigger display for the old results before updateDropdown finish executing a new search)

cea2aj: With this approach, some callbacks are async and other are not, and that distinction might seem arbitrary to a consumer of this component. I think the simplest solution would be to update the code inside the `onClick` handler to use the `.then` syntax (e.g. `updateDropdown().then(...` so that we can avoid making this async. Alternatively we might want to consider supporting both async and non-async callbacks (I'd want to have a larger discussion with Tom before going this route)

yen-tt: hm I see. updateDropdown would have to return Promise<void> in order to use the `.then` syntax. Perhaps we can make a distinction with the naming between the callbacks that requires network calls and those that's just state modification like tracking input string? I will talk to tom if we want to support non-async callbacks as well

oshi97: would it make sense to move this to a different PR?

oshi97: actually ignore me it's probably best to just resolve this here

sample-app/src/components/SearchBar.tsx
```diff
@@ -19,17 +20,20 @@ interface Props {
...
-    : state => state.universal.autoComplete?.results;
-  const autocompleteResults = useAnswersState(mapStateToAutocompleteResults) || [];
-  const isLoading = useAnswersState(state => state.vertical.searchLoading || state.universal.searchLoading);
-
-  function executeAutocomplete () {
+  const query = useAnswersState(state => state.query.input);
+  const [ autocompleteResults, setAutoCompleteResults ] = useState<AutocompleteResult[]>([]);
+  const isLoading = useAnswersState(state => state.searchStatus.isLoading);
+  const autocompleteNetworkIds = useRef({ latestRequest: 0, responseInState: 0 });
+  async function executeAutocomplete () {
```
oshi97, line 27: could we make a useAutocomplete hook? I was thinking of something like the below, it wouldn't go into the react bindings (at least not yet) but just live in some sample-app/src/hooks/ folder
```ts
function useAutocomplete(isVertical) {
  const autocompleteNetworkIds = useRef({ latestRequest: 0, responseInState: 0 });
  const [ autocompleteResults, setAutoCompleteResults ] = useState<AutocompleteResult[]>([]);
  async function executeAutocomplete () {
    // what you have
  }
  return [ autocompleteResults, executeAutocomplete ]
}
```

oshi97: I think it's ok to make it very specific to autocomplete for now, later on if we need we can make a more general hook that just does the network request race condition handling

yen-tt: yeah, makes sense

sample-app/src/components/InputDropdown.tsx
yen-tt: removed the async stuff for now! will add support for non-async and async callbacks in another item

sample-app/src/hooks/useAutocomplete.tsx
```diff
@@ -0,0 +1,20 @@
...
+import { AutocompleteResult, useAnswersActions } from '@yext/answers-headless-react';
+
+export function useAutocomplete(isVertical: boolean): [AutocompleteResult[], () => Promise<void>] {
+  const answersActions = useAnswersActions();
+  const autocompleteNetworkIds = useRef({ latestRequest: 0, responseInState: 0 });
+  const [ autocompleteResults, setAutoCompleteResults ] = useState<AutocompleteResult[]>([]);
+  async function executeAutocomplete () {
+    const requestId = ++autocompleteNetworkIds.current.latestRequest;
+    let response = null;
+    isVertical
```
oshi97, line 11: is it possible to do this with a `const response`?

yen-tt: oh yeah

sample-app/src/hooks/useAutocomplete.tsx
```diff
@@ -0,0 +1,19 @@
+import { useRef, useState } from "react";
+import { AutocompleteResult, useAnswersActions } from '@yext/answers-headless-react';
+
+export function useAutocomplete(isVertical: boolean): [AutocompleteResult[], () => Promise<void>] {
+  const answersActions = useAnswersActions();
+  const autocompleteNetworkIds = useRef({ latestRequest: 0, responseInState: 0 });
+  const [ autocompleteResults, setAutoCompleteResults ] = useState<AutocompleteResult[]>([]);
```
oshi97, line 7: we have an item to be consistent about the capitzliation for AutoComplete vs Autocomplete but just making a note here

yen-tt: oh yea, good catch. I made some changes to headless to have it lowercase, missed it here

oshi97 (approved):

PR #55, opened 2021-11-16
fix testRunner in package.json

oshi97:
This path does not exist anymore even on my local machine.

J=none
TEST=manual

can do "npm run test"

### DIFF
--- a/sample-app/package.json
+++ b/sample-app/package.json
@@ -134,7 +134,7 @@
       "<rootDir>/src/**/*.{spec,test}.{js,jsx,ts,tsx}"
     ],
     "testEnvironment": "jsdom",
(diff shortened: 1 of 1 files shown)


sample-app/package.json
```diff
@@ -134,7 +134,7 @@
       "<rootDir>/src/**/*.{spec,test}.{js,jsx,ts,tsx}"
     ],
     "testEnvironment": "jsdom",
-    "testRunner": "/Users/oshi/answers-headless/react-test-site/node_modules/jest-circus/runner.js",
+    "testRunner": "./node_modules/jest-circus/runner.js",
```
cea2aj, line 137: I know we're not using this, but should it be `<rootDir>` instead?

oshi97: either works (also tested cd-ing into a nested folder)

cea2aj (approved):

PR #56, opened 2021-11-17
Add FilterSearch component

nmanu1:
Create a component for FilterSearch. Update InputDropdown and Dropdown to have sections of dropdown options.

J=SLAP-1643
TEST=manual

Conduct filter searches in sample-app to check the behavior is as expected. Check that search bar dropdown functions as before.

### DIFF
--- a/sample-app/src/components/Dropdown.tsx
+++ b/sample-app/src/components/Dropdown.tsx
@@ -34,15 +34,13 @@ export default function Dropdown({
   optionIdPrefix,
   cssClasses
 }: Props): JSX.Element | null {
-  function renderSection(section: {results: Option[], label?: string}, sectionIndex: number) {
+  function renderSection(section: { results: Option[], label?: string }, sectionIndex: number) {
     return (
       <div
         className={cssClasses.optionSection}
         key={`section-${sectionIndex}`}>
         {section.label &&
-          <div
-            className={cssClasses.sectionLabel}
-            key={section.label}>
+          <div className={cssClasses.sectionLabel}>
             {section.label}
           </div>
         }
--- a/sample-app/src/components/Dropdown.tsx
+++ b/sample-app/src/components/Dropdown.tsx
@@ -6,12 +6,15 @@ export interface Option {
 }
 
 interface Props {
-  options: Option[],
-  onClickOption?: (option: Option) => void,
+  options: {results: Option[], label?: string}[],
+  onClickOption?: (option: Option, sectionIndex: number, optionIndex: number) => void,
+  focusedSectionIndex: number | undefined,
   focusedOptionIndex: number | undefined,
   optionIdPrefix: string,
   cssClasses: {
     optionContainer: string,
+    optionSection: string,
+    sectionLabel: string,
     option: string,
     focusedOption: string
   }
@@ -23,22 +26,41 @@ interface Props {
 export default function Dropdown({
   options,
   onClickOption = () => {},
+  focusedSectionIndex,
   focusedOptionIndex,
   optionIdPrefix,
   cssClasses
 }: Props): JSX.Element | null {
-  function renderOption(option: Option, index: number) {
+  function renderSection(section: {results: Option[], label?: string}, sectionIndex: number) {
+    return (
+      <div
+        className={cssClasses.optionSection}
+        key={`section-${sectionIndex}`}>
+        {section.label &&
+          <div
+            className={cssClasses.sectionLabel}
+            key={section.label}>
+            {section.label}
+          </div>
+        }
+        {section.results.map((option, optionIndex) => renderOption(option, sectionIndex, optionIndex))}
+      </div>
+    )
+  }
+
+  function renderOption(option: Option, sectionIndex: number, optionIndex: number) {
     const className = classNames(cssClasses.option, {
-      [cssClasses.focusedOption]: index === focusedOptionIndex
+      [cssClasses.focusedOption]: sectionIndex === focusedSectionIndex && optionIndex === focusedOptionIndex
     })
     return (
       <div 
-        key={index}
+        key={`${sectionIndex}-${optionIndex}`}
         className={className}
-        id={`${optionIdPrefix}-${index}`}
-        onClick={() => onClickOption(option)}>
+        id={`${optionIdPrefix}-${sectionIndex}-${optionIndex}`}
+        onClick={() => onClickOption(option, sectionIndex, optionIndex)}>
         {option.render()}
-      </div>)
+      </div>
+    )
   }
 
   if (options.length < 1) {
@@ -47,7 +69,7 @@ export default function Dropdown({
 
   return (
     <div className={cssClasses.optionContainer}>
-      {options.map((option, index) => renderOption(option, index))}
+      {options.map((section, sectionIndex) => renderSection(section, sectionIndex))}
     </div>
   );
 };
--- a/sample-app/src/components/Dropdown.tsx
+++ b/sample-app/src/components/Dropdown.tsx
@@ -70,7 +70,7 @@ export default function Dropdown({
 
   return (
     <div className={cssClasses.optionContainer}>
-      {options.map((section, sectionIndex) => renderSection(section, sectionIndex))}
+      {options.map(renderSection)}
     </div>
   );
 };
--- a/sample-app/src/components/Dropdown.tsx
+++ b/sample-app/src/components/Dropdown.tsx
@@ -6,7 +6,10 @@ export interface Option {
 }
 
 interface Props {
-  options: {results: Option[], label?: string}[],
+  options: {
+    results: Option[], 
+    label?: string
+  }[],
   onClickOption?: (option: Option, sectionIndex: number, optionIndex: number) => void,
   focusedSectionIndex: number | undefined,
   focusedOptionIndex: number | undefined,
--- a/sample-app/src/components/FilterSearch.tsx
+++ b/sample-app/src/components/FilterSearch.tsx
@@ -1,6 +1,6 @@
 import { FilterSearchResponse } from "@yext/answers-core";
 import { SearchParameterField } from "@yext/answers-core";
-import { useCallback, useEffect, useRef, useState } from "react";
+import { useRef, useState } from "react";
 import { useAnswersActions } from '@yext/answers-headless-react';
 import InputDropdown from "./InputDropdown";
 import renderWithHighlighting from "./utils/renderWithHighlighting";
@@ -15,74 +15,72 @@ export interface FilterSearchProps {
   screenReaderInstructionsId: string
 }
 
-
 export default function FilterSearch (props: FilterSearchProps): JSX.Element {
   const { title, sectioned, searchFields, screenReaderInstructionsId } = props;
   const answersActions = useAnswersActions();
-  const input = useRef<string>('');
+  const [input, setInput] = useState('');
   const [results, updateResults] = useState<FilterSearchResponse|undefined>();
   const [, setMessage] = useState( '' );
   const requestId = useRef(0);
   const responseId = useRef(0);
 
-
-
-  const executeSearch = useCallback(async () => {
+  async function executeFilterSearch (inputValue: string) {
     const currentId = ++requestId.current;
-    const results = await answersActions.executeFilterSearch(input.current, sectioned, searchFields);
+    const results = await answersActions.executeFilterSearch(inputValue, sectioned, searchFields);
     if (currentId >= responseId.current) {
       responseId.current++;
       updateResults(results);
     }
-  }, [answersActions, searchFields, sectioned]);
+  }
 
-  useEffect(() => {
-    executeSearch();
-  }, [executeSearch]);
-
-  let stuff: Option[]  = [];
+  let options: {results: Option[], label?: string}[] = [];
   if (results) {
-    if(results.sectioned) {
-      results.sections.forEach(section => {
-        section.results.forEach(result => {
-          stuff.push({
-            value: result.value,
-            render: () => renderWithHighlighting(result)
-          });
-        })
-      });
-    } else {
-      stuff = results.results.map(result => {
-        return {
+    results.sections.forEach(section => {
+      let results: Option[] = [];
+      section.results.forEach(result => {
+        results.push({
           value: result.value,
           render: () => renderWithHighlighting(result)
-        }
+        });
       });
-    }
+      options.push({
+        results: results,
+        label: section.label
+      });
+    });
   }
 
   return (
     <div className='FilterSearch'>
       <h1>{title}</h1>
       <InputDropdown
-        inputValue={input.current}
+        inputValue={input}
         placeholder='this is filter search...'
         screenReaderInstructions={SCREENREADER_INSTRUCTIONS}
         screenReaderInstructionsId={screenReaderInstructionsId}
-        options={stuff}
+        options={options}
         optionIdPrefix='Autocomplete__option'
-        onSubmit={() => {
-          answersActions.executeVerticalQuery();
-          setMessage('enter!');
+        onlySubmitOnOption={true}
+        onSubmit={(optionValue, sectionIndex, optionIndex) => {
+          if (sectionIndex !== undefined && optionIndex !== undefined && results) {
+            const option = results.sections[sectionIndex].results[optionIndex];
+            if (option.filter) {
+              answersActions.setFilterOption({ ...option.filter, selected: true }, 'someFiltersSearchId');
+              answersActions.executeVerticalQuery();
+              setMessage('enter!');
+            }
+          }
         }}
         updateInputValue={newInput => {
-          input.current = newInput;
+          setInput(newInput);
         }}
-        updateDropdown={() => {
-          executeSearch();
+        updateDropdown={(input) => {
+          executeFilterSearch(input);
         }}
         cssClasses={{
           optionContainer: 'Autocomplete',
+          optionSection: 'Autocomplete__optionSection',
+          sectionLabel: 'Autocomplete__sectionLabel',
           option: 'Autocomplete__option',
           focusedOption: 'Autocomplete__option--focused',
           inputElement: 'FilterSearch__input',
--- a/sample-app/src/components/FilterSearch.tsx
+++ b/sample-app/src/components/FilterSearch.tsx
@@ -37,18 +37,16 @@ export default function FilterSearch (props: FilterSearchProps): JSX.Element {
 
   let options: { results: Option[], label?: string }[] = [];
   if (results) {
-    results.sections.forEach(section => {
-      let results: Option[] = [];
-      section.results.forEach(result => {
-        results.push({
-          value: result.value,
-          render: () => renderWithHighlighting(result)
-        });
-      });
-      options.push({
-        results: results,
+    options = results.sections.map(section => {
+      return {
+        results: section.results.map(result => {
+          return {
+            value: result.value,
+            render: () => renderWithHighlighting(result)
+          };
+        }),
         label: section.label
-      });
+      };
     });
   }
 
--- a/sample-app/src/components/FilterSearch.tsx
+++ b/sample-app/src/components/FilterSearch.tsx
@@ -65,7 +65,7 @@ export default function FilterSearch (props: FilterSearchProps): JSX.Element {
           if (sectionIndex !== undefined && optionIndex !== undefined && results) {
             const option = results.sections[sectionIndex].results[optionIndex];
             if (option.filter) {
-              answersActions.setFilterOption({ ...option.filter, selected: true }, 'someFiltersSearchId');
+              answersActions.setFilterOption({ ...option.filter, selected: true });
               answersActions.executeVerticalQuery();
               setMessage('enter!');
             }
--- a/sample-app/src/components/FilterSearch.tsx
+++ b/sample-app/src/components/FilterSearch.tsx
@@ -1,0 +1,94 @@
+import { FilterSearchResponse } from "@yext/answers-core";
+import { SearchParameterField } from "@yext/answers-core";
+import { useCallback, useEffect, useRef, useState } from "react";
+import { useAnswersActions } from '@yext/answers-headless-react';
+import InputDropdown from "./InputDropdown";
+import renderWithHighlighting from "./utils/renderWithHighlighting";
+import { Option } from "./Dropdown";
+
+const SCREENREADER_INSTRUCTIONS = 'When autocomplete results are available, use up and down arrows to review and enter to select.'
+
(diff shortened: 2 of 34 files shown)


sample-app/src/components/FilterSearch.tsx
```diff
@@ -0,0 +1,92 @@
...
+export default function FilterSearch (props: FilterSearchProps): JSX.Element {
+  const { title, sectioned, searchFields, screenReaderInstructionsId } = props;
+  const answersActions = useAnswersActions();
+  const [input, setInput] = useState('');
+  const [results, updateResults] = useState<FilterSearchResponse|undefined>();
+  const [, setMessage] = useState( '' );
+  const requestId = useRef(0);
+  const responseId = useRef(0);
+
+  async function executeFilterSearch (inputValue: string) {
```
yen-tt, line 27: I think we can use the `useAutocomplete` hook but maybe call it `useSynchronizedSearch` (probably a better name) and takes in a function to execute either autocomplete or filtersearch.

sample-app/src/components/FilterSearch.tsx
```diff
@@ -0,0 +1,92 @@
...
+        options={options}
+        optionIdPrefix='Autocomplete__option'
+        onlySubmitOnOption={true}
+        onSubmit={(optionValue, sectionIndex, optionIndex) => {
+          if (sectionIndex !== undefined && optionIndex !== undefined && results) {
+            const option = results.sections[sectionIndex].results[optionIndex];
+            if (option.filter) {
+              answersActions.setFilterOption({ ...option.filter, selected: true });
+              answersActions.executeVerticalQuery();
+              setMessage('enter!');
```
yen-tt, line 70: forgot to remove?

sample-app/src/components/FilterSearch.tsx
```diff
@@ -0,0 +1,92 @@
...
+  searchFields: SearchParameterField[],
+  screenReaderInstructionsId: string
+}
+
+export default function FilterSearch (props: FilterSearchProps): JSX.Element {
+  const { title, sectioned, searchFields, screenReaderInstructionsId } = props;
+  const answersActions = useAnswersActions();
+  const [input, setInput] = useState('');
+  const [results, updateResults] = useState<FilterSearchResponse|undefined>();
+  const [, setMessage] = useState( '' );
```
yen-tt, line 23: I think this is can be remove

sample-app/src/components/FilterSearch.tsx
```diff
@@ -0,0 +1,92 @@
+import { FilterSearchResponse } from "@yext/answers-core";
+import { SearchParameterField } from "@yext/answers-core";
```
yen-tt, line 2: I think you import these stuff through @yext/answers-headless-react

sample-app/src/components/FilterSearch.tsx
nmanu1: oops

sample-app/src/components/Dropdown.tsx
```diff
@@ -23,22 +26,41 @@ interface Props {
 export default function Dropdown({
   options,
   onClickOption = () => {},
+  focusedSectionIndex,
   focusedOptionIndex,
   optionIdPrefix,
   cssClasses
 }: Props): JSX.Element | null {
-  function renderOption(option: Option, index: number) {
+  function renderSection(section: {results: Option[], label?: string}, sectionIndex: number) {
```
yen-tt, line 34: nit: spacing inside the curly braces. hmm I thought `"object-curly-spacing": ["error", "always"]` would catch this. Can you double check if that eslint rule is working?

sample-app/src/components/Dropdown.tsx
```diff
@@ -6,12 +6,15 @@ export interface Option {
 }
 
 interface Props {
-  options: Option[],
-  onClickOption?: (option: Option) => void,
+  options: {results: Option[], label?: string}[],
```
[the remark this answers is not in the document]
nmanu1: updated!

sample-app/src/components/FilterSearch.tsx
yen-tt: ah yep

sample-app/src/components/InputDropdown.tsx
```diff
@@ -80,9 +87,36 @@ export default function InputDropdown({
...
+      phrase: `0 autocomplete option found.`,
+      pluralForm: `0 autocomplete options found.`,
+      count: 0
+    })
+    screenReaderPhrases.push(phrase);
+  } else {
+    options.forEach(section => {
+      const phrase = section.label ?
+        processTranslation({
+          phrase: `${section.results.length} ${section.label} autocomplete option found.`,
```
yen-tt, line 102: can we group these two process translation into one? like either construct `${section.results.length} ${section.label}` outside then call processTranslation or check the label directly in string and add '' if it doesn't exist?

sample-app/src/components/Dropdown.tsx
```diff
@@ -23,22 +29,41 @@ interface Props {
...
-  function renderOption(option: Option, index: number) {
+  function renderSection(section: {results: Option[], label?: string}, sectionIndex: number) {
+    return (
+      <div
+        className={cssClasses.optionSection}
+        key={`section-${sectionIndex}`}>
+        {section.label &&
+          <div
+            className={cssClasses.sectionLabel}
+            key={section.label}>
```
cea2aj, line 45: Should we have `sectionIndex` as part of the key as well in case two sections have the same label? It's unlikely, but adding more to the key shouldn't hurt

sample-app/src/components/InputDropdown.tsx
```diff
@@ -97,24 +131,43 @@ export default function InputDropdown({
...
     const isFirstOptionFocused = focusedOptionIndex === 0;
-    const isLastOptionFocused = focusedOptionIndex === options.length - 1;
+    const isLastOptionFocused = 
+      focusedSectionIndex !== undefined &&
+      focusedOptionIndex === options[focusedSectionIndex].results.length - 1;
     if (evt.key === 'Enter') {
-      updateInputValue(inputValue);
-      onSubmit(inputValue);
-      dispatch({ type: 'HideOptions' });
+      if (!onlySubmitOnOption || focusedOptionIndex !== undefined) {
```
yen-tt, line 141: what is onlySubmitOnOption, and why do we need it?

sample-app/src/components/InputDropdown.tsx
```diff
@@ -97,24 +131,43 @@ export default function InputDropdown({
...
+        onSubmit(inputValue, focusedSectionIndex, focusedOptionIndex);
+        dispatch({ type: 'HideOptions' });
+      }
     } else if (evt.key === 'Escape' || evt.key === 'Tab') {
       dispatch({ type: 'HideOptions' });
-    } else if (evt.key === 'ArrowDown' && options.length > 0 && !isLastOptionFocused) {
-      const newIndex = focusedOptionIndex !== undefined ? focusedOptionIndex + 1 : 0;
-      dispatch({ type: 'FocusOption', newIndex });
-      const newValue = options[newIndex]?.value;
+    } else if (evt.key === 'ArrowDown' && options.length > 0 && !(isLastSectionFocused && isLastOptionFocused)) {
```
yen-tt, line 147: do we still need to check for `options.length > 0` if we also check for !(isLastSectionFocused && isLastOptionFocused)?

sample-app/src/components/InputDropdown.tsx
nmanu1: when onlySubmitOnOption is true, the input should not be submitted when 'enter' is pressed, unless it is focused on an option. otherwise, if onlySubmitOnOption is false, when 'enter' is pressed, the user input is submitted if no option is focused.

if we dispatch `HideOptions` regardless, then if a user types something into the input bar for filter search and presses 'enter' without focusing on a filter option, the dropdown would be hidden but no filter would be applied. is that the behavior we want, or would we want to keep the dropdown open?

sample-app/src/pages/VerticalSearchPage.tsx
```diff
@@ -54,6 +55,17 @@ const staticFiltersGroupLabels = {
   c_employeeCountry: 'Employee Country'
 }
 
+const filterSearchFields = [{
+  fieldApiName: 'builtin.location',
+  entityType: 'ce_person',
+  fetchEntities: false
```
cea2aj, line 61: If we enable `fetchEntities`, it won't affect the FilterSearch component at all, correct? Because of that, I'm thinking we shouldn't have `fetchEntities` as an option for the component. We could use a Typescript utility type to omit the field, and inside the component we can always set it to false

sample-app/src/components/InputDropdown.tsx
yen-tt: I see, can you check what does sdk do? I think it makes sense to close it if user press enter without focusing on an option

sample-app/src/components/InputDropdown.tsx
nmanu1: yes, because otherwise if both focused indexes are undefined, !(isLastSectionFocused && isLastOptionFocused) would be true, and it would try to arrow down to an option that doesn't exist.

sample-app/src/components/InputDropdown.tsx
nmanu1: the SDK leaves it open

sample-app/src/components/Dropdown.tsx
nmanu1: yes, I'll add that

nmanu1: actually, the key is not needed here. that was just from a previous version, so I'll remove it entirely

sample-app/src/components/Dropdown.tsx
```diff
@@ -23,22 +29,41 @@ interface Props {
...
   options,
   onClickOption = () => {},
+  focusedSectionIndex,
   focusedOptionIndex,
   optionIdPrefix,
   cssClasses
 }: Props): JSX.Element | null {
-  function renderOption(option: Option, index: number) {
+  function renderSection(section: {results: Option[], label?: string}, sectionIndex: number) {
+    return (
```
cea2aj, line 38: Do we want to enforce that each section must consist of an array of results? We could make it more generic by allowing an array of sections to be passed in which define their own render function similarly to how options can be passed in. I believe Tom mentioned a use case such as Visual Autocomplete where we'd want sections, but they wouldn't necessarily be laid out this way. Have you checked with Tom on the use cases here?

sample-app/src/components/InputDropdown.tsx
```diff
@@ -8,14 +8,17 @@ interface Props {
   placeholder?: string,
   screenReaderInstructions: string,
   screenReaderInstructionsId: string,
-  options: Option[],
+  options: {results: Option[], label?: string}[],
```
cea2aj, line 11: This changes forcers consumers of this component to organize the dropdown into sections. Is there a way we could avoid that?

nmanu1: The only way I could think of to avoid that would be to keep the options array as is, but pass in an optional array of strings and numbers for the labels of each section and the number of results in each section. This would assume that the options array is ordered by section in the same order as the labels and counts.

Something like `additionalData?: {labels: string[], counts: number[]}` or 
`additionalData?: {label: string, count: number}[]`

sample-app/src/components/FilterSearch.tsx
cea2aj: abstracting out this logic so that we don't duplicate it makes sense to me

sample-app/src/components/InputDropdown.tsx
yen-tt: another way would be allowing for both sections and the previous array of options 
`options: {results: Option[], label?: string}[] | Option[]`
and either InputDropdown support both logic, or transform options to a section internally if it's not one

sample-app/src/components/Dropdown.tsx
nmanu1: Added the spaces, but seems like the eslint rule isn't working because it didn't give any error when there were no spaces

yen-tt: Can you do a brief check on why that is? it seems to work for the top level src folder outside of sample-app

sample-app/src/components/InputDropdown.tsx
yen-tt: I would imagine that `processTranslation` will eventually work like the same function we have in SDK use for  translations, which includes interpolation and pluralization. Maybe we can keep the call for no results

sample-app/src/components/InputDropdown.tsx
yen-tt: can we default the indexes to 0?

sample-app/src/components/SearchBar.tsx
```diff
@@ -59,13 +59,16 @@ export default function SearchBar({ placeholder, isVertical, screenReaderInstruc
...
-          }
-        })}
+        options={autocompleteResults.length > 0 ? [{
+          results: autocompleteResults.map(result => {
+            return {
+              value: result.value,
+              render: () => renderWithHighlighting(result)
+            }
+          })
+        }] : []}
```
yen-tt, line 60: can we simplify to without the ternary operator?

sample-app/src/components/InputDropdown.tsx
```diff
@@ -97,24 +131,43 @@ export default function InputDropdown({
...
     }
 
+    const isFirstSectionFocused = focusedSectionIndex === 0;
+    const isLastSectionFocused = focusedSectionIndex === options.length - 1;
     const isFirstOptionFocused = focusedOptionIndex === 0;
-    const isLastOptionFocused = focusedOptionIndex === options.length - 1;
+    const isLastOptionFocused = 
+      focusedSectionIndex !== undefined &&
+      focusedOptionIndex === options[focusedSectionIndex].results.length - 1;
     if (evt.key === 'Enter') {
```
yen-tt, line 130: can we combined these two ifs since we don't do anything specific for just enter now?

sample-app/src/components/InputDropdown.tsx
yen-tt: gotcha

sample-app/src/components/InputDropdown.tsx
cea2aj: How about keeping the options data the same but making it optional, and adding an optional sections param with a different interface?

sample-app/src/components/InputDropdown.tsx
```diff
@@ -97,24 +121,43 @@ export default function InputDropdown({
       evt.preventDefault();
     }
 
+    const isFirstSectionFocused = focusedSectionIndex === 0;
```
cea2aj, line 124: Adding sections support is adding a lot of complexity to this logic. If we support either an array of options or a sections object, then we could have separate functions where one handles the logic for simple options, and the other handles the logic for sections.

sample-app/src/components/InputDropdown.tsx
yen-tt: would it seems a little odd if there's two params for passing options?

sample-app/src/components/FilterSearch.tsx
```diff
@@ -0,0 +1,90 @@
...
+    const results = await answersActions.executeFilterSearch(inputValue, sectioned, searchFields);
+    if (currentId >= responseId.current) {
+      responseId.current++;
+      updateResults(results);
+    }
+  }
+
+  let options: {results: Option[], label?: string}[] = [];
+  if (results) {
+    results.sections.forEach(section => {
```
oshi97, line 37: does flatMap work here?

sample-app/src/components/Dropdown.tsx
```diff
@@ -47,7 +70,7 @@ export default function Dropdown({
 
   return (
     <div className={cssClasses.optionContainer}>
-      {options.map((option, index) => renderOption(option, index))}
+      {options.map((section, sectionIndex) => renderSection(section, sectionIndex))}
```
oshi97, line 73: nit but I think just `{options.map(renderSection}` should work here? no need for extra anonymous function

sample-app/src/components/Dropdown.tsx
```diff
@@ -23,22 +29,39 @@ interface Props {
 export default function Dropdown({
   options,
   onClickOption = () => {},
+  focusedSectionIndex,
   focusedOptionIndex,
   optionIdPrefix,
   cssClasses
 }: Props): JSX.Element | null {
-  function renderOption(option: Option, index: number) {
+  function renderSection(section: { results: Option[], label?: string }, sectionIndex: number) {
```
oshi97, line 37: I might consider moving these functions to the top level of the file. I don't think it matters in this case, but react uses reference equality in a number of places, and here new `renderSection()` and `renderOption()` functions are used for each render of `Dropdown`. If we move them to the top level the functions will have stable references. AGAIN I don't think matters here because we are not passing these functions in as props anywhere or using them in as memo dependencies etc. but I also don't have a compiler in my brain

sample-app/src/components/FilterSearch.tsx
```diff
@@ -0,0 +1,90 @@
...
+          }
+        }}
+        updateInputValue={newInput => {
+          setInput(newInput);
+        }}
+        updateDropdown={(input) => {
+          executeFilterSearch(input);
+        }}
+        cssClasses={{
+          optionContainer: 'Autocomplete',
```
oshi97, line 79: Would it be possible to make all of the "Autocomplete" prefixed css classes the default values? Or maybe we don't even need props for them, at least not yet?.

sample-app/src/components/SearchBar.tsx
nmanu1: I added the ternary operator because without it, it would pass `[{ results: [] }]` as `options`, which results in `options.length > 0` checks incorrectly passing

sample-app/src/components/FilterSearch.tsx
nmanu1: I was going off of the css classes in the SearchBar component, but I can remove the props from both components, or change them to be optional with the `Autocomplete` prefixed classes as the default in the Dropdown component. It would just depend on whether we want to require or even support custom css class names for that

sample-app/src/components/Dropdown.tsx
nmanu1: I looked into it and seems like the sample-app folder is one of the ignoredPatterns in the eslint config. Should I remove it from there so that it gives the errors?

sample-app/src/components/SearchBar.tsx
yen-tt: I see that filterSearch, on a no results would have `sections: [{results: []}]` in the network response, so the filterSearch component right now would have `[{ results: [] }]` options. That's working?

nmanu1: When there are no results found, executeFilterSearch returns `sections: []` so it passes `options: []` to the InputDropdown

sample-app/src/components/InputDropdown.tsx
nmanu1: sure!

sample-app/src/components/FilterSearch.tsx
oshi97: ah yeah I'm kind of scope creeping here, we can keep the classes for now if you like (also Connor is doing some css stuff right now so maybe that's for the best)

sample-app/src/components/InputDropdown.tsx
nmanu1: Makes sense. I added it back for no results

sample-app/src/pages/VerticalSearchPage.tsx
nmanu1: Updated!

sample-app/src/components/FilterSearch.tsx
nmanu1: ok, I think I'll keep it as is for now then

sample-app/src/components/Dropdown.tsx
yen-tt: inside package.json there's an eslintConfig part for sample-app that's different from the top level eslint. So I believe the ignorePatterns stuff shouldn't override the internal rules for sample-app

nmanu1: I didn't change anything, so not sure why, but now I'm getting the eslint errors in the sample-app

sample-app/src/components/InputDropdown.tsx
```diff
@@ -80,9 +87,31 @@ export default function InputDropdown({
...
+      });
+      screenReaderPhrases.push(phrase);
+    });
+  }
+  const screenReaderText = screenReaderPhrases.join(' ');
+
   function handleDocumentClick(evt: MouseEvent) {
     const target = evt.target as HTMLElement;
-    if (!target.isSameNode(inputRef.current)) {
+    if (!(target.isSameNode(inputRef.current) || target.classList.contains(cssClasses.optionContainer)
```
cea2aj, line 113: The cssClasses object may contain a list of classes, so classList.contains wouldn't work in that scenario. Can we check in some other way that doesn't rely on the classList?

sample-app/src/components/Dropdown.tsx
nmanu1: good catch! updated

sample-app/src/components/FilterSearch.tsx
nmanu1: I changed it to use map instead

sample-app/src/components/InputDropdown.tsx
nmanu1: The other way I thought of would involve useRef and forwardRef, but I think there would need to be a reference for each section label

sample-app/src/components/InputDropdown.tsx
```diff
@@ -97,24 +124,41 @@ export default function InputDropdown({
...
       dispatch({ type: 'HideOptions' });
     } else if (evt.key === 'Escape' || evt.key === 'Tab') {
       dispatch({ type: 'HideOptions' });
-    } else if (evt.key === 'ArrowDown' && options.length > 0 && !isLastOptionFocused) {
-      const newIndex = focusedOptionIndex !== undefined ? focusedOptionIndex + 1 : 0;
-      dispatch({ type: 'FocusOption', newIndex });
-      const newValue = options[newIndex]?.value;
+    } else if (evt.key === 'ArrowDown' && options.length > 0 && !(isLastSectionFocused && isLastOptionFocused)) {
+      let newSectionIndex, newOptionIndex;
+      if (isLastOptionFocused) {
```
oshi97, line 140: would it make sense to move some of this computation to the reducer?

PR #57, opened 2021-11-17
Support "Near Me" queries

yen-tt:
Add support for 'Near me' queries in search bar

- added `search-operations` file that includes util functions for executing search, getting search intents, and user location
- update LocationBias component to retrieve user location and execute search using SearchHandler.
- updated `VerticalSearchPage` and `UniversalSearchPage` fetch searchIntents from autocomplete search, in the case of a default initial search with near me intent, get and update user's location if there is a near me intent, and execute query.
- updated searchBar to rely on autocomplete response from latest request to retrieve the search intent, and update user's location before executing a search

Note: there's a noticeable wait time when navigator.geolocation is used. and there isn't any indication that the results maybe  in loading state. Should consider exposing a setLoadingStatus in answersHeadless, so it can be use when executeSearchWithUserLocation is triggered. 

J=SLAP-1706
TEST=manual

- set location to washington DC in when vertical page is loaded, click update my location from `LocationBias` component, and see that the label is updated to my device's location.
- set default search query to 'engineers near me' and see that user location is set based on location near my device.
- ran searches between non-nearMe and nearMe queries, see that setUserLocation is only on queries with nearMe intents


sample-app/src/components/LocationBias.tsx
```diff
@@ -1,12 +1,9 @@
...
+import { executeSearchWithUserLocation } from '../utils/geolocationutils';
 
 interface Props {
   isVertical: boolean,
-  geolocationOptions?: {
-    enableHighAccuracy?: boolean,
-    timeout?: number,
-    maximumAge?: number
-  },
+  geolocationOptions?: PositionOptions,
```
tmeyer2115, line 6: Where is the `PositionOptions` type coming from? Are we missing an import or is this supposed to be defined in the file?

sample-app/src/components/SearchBar.tsx
```diff
@@ -25,9 +26,7 @@ export default function SearchBar({ placeholder, isVertical, screenReaderInstruc
   const isLoading = useAnswersState(state => state.searchStatus.isLoading);
 
   function executeQuery () {
-    isVertical
-      ? answersActions.executeVerticalQuery()
-      : answersActions.executeUniversalQuery();
+    executeSearchWithUserLocation(answersActions, isVertical, {}, true);
```
tmeyer2115, line 29: This reads a bit funny. To a developer, it would seem as if we're always including the userLocation in the search. Additionally, the SDK's `SearchBar` allows you to configure the GeoLocation call. So, we should probably add `PositionProps` to `Props` here too.

sample-app/src/utils/geolocationutils.tsx
```diff
@@ -0,0 +1,56 @@
+import { AnswersActions } from "@yext/answers-headless-react";
+import { SearchIntent } from "@yext/answers-headless";
+
+const defaultGeolocationOptions: PositionOptions = {
```
tmeyer2115, line 4: Is `PositionOptions` some built-in type? I don't see it getting defined or imported anywhere.

sample-app/src/utils/geolocationutils.tsx
```diff
@@ -0,0 +1,56 @@
...
+ * Executes universal or vertical search with user location retrieved from nagivator.geolocation API.
+ * If useLocationOnNearMeOnly is true, will attempt to get user location only if the query's searchIntents
+ * (retrieved through autocomplete response) contains 'NEAR_ME'. If there's a failure in getting
+ * user location, a normal search is performed.
+ */
+export async function executeSearchWithUserLocation(
+  answersActions: AnswersActions,
+  isVertical: boolean,
+  geolocationOptions?: PositionOptions,
+  useLocationOnNearMeOnly?: boolean
```
tmeyer2115, line 20: I think we should consider renaming this method. The method name suggests we always provide the user's location. But, if you dig deeper, you see there's a param that lets us search without sending the user's location.

sample-app/src/utils/geolocationutils.tsx
```diff
@@ -0,0 +1,56 @@
...
+    },
+    (err) => {
+      console.error(err);
+      console.error('Unable to determine user\'s location.');
+    },
+    Object.assign(defaultGeolocationOptions, geolocationOptions)
+    );
+  } else {
+    console.warn('Unable to determine user\'s location.');
+  }
```
tmeyer2115, line 54: This function is doing a lot. It might be worth considering turning this into some sore of class, such as a `SearchHandler`. We may want to use a behavioral design pattern here. I'm trying to think which one might be most appropriate.

sample-app/src/utils/geolocationutils.tsx
yen-tt: yup, it's a built-in type

sample-app/src/components/LocationBias.tsx
yen-tt: yea, `PositionOptions` and `GeolocationPosition` are both built-in types

sample-app/src/utils/geolocationutils.tsx
tmeyer2115: Specifically, I think we should try and decompose what's going on in here. There are multiple responsibilities in this method: fire off a search, see if there's near me intent, and get the user's location. This might be total overkill, but we could use the `Command` pattern and generate three commands:
- SearchCommand
- FetchLocationCommand
- FetchSearchIntentsCommand

Those aren't necessarily the best names. Just an idea, but I think we should consider changing this up to be more than just a single complex method.

sample-app/src/utils/geolocationutils.tsx
```diff
@@ -0,0 +1,56 @@
...
+      ? answersActions.executeVerticalQuery()
+      : answersActions.executeUniversalQuery();
+    return;
+  }
+
+  if (useLocationOnNearMeOnly) {
+    const results = isVertical
+      ? await answersActions.executeVerticalAutocomplete()
+      : await answersActions.executeUniversalAutocomplete();
+    if (!results?.inputIntents?.includes(SearchIntent.NearMe)) {
```
oshi97, line 33: as somebody is typing queries into the searchbar, each key stroke will already fire off an autocomplete request. Would it be possible to use the inputIntents from there instead?

yen-tt: that's not stored in headless state, but rather inside searchbar component state. it would be tricky to pass that around to different components. Maybe the searchIntent of autocomplete should be store somewhere, or have the ability to update the searchIntent in QueryState? what do you think

sample-app/src/utils/geolocationutils.tsx
```diff
@@ -0,0 +1,56 @@
+import { AnswersActions } from "@yext/answers-headless-react";
```
oshi97, line 1: I don't think this file needs to be .tsx and can just be .ts. Maybe we could change the other "blah-utils.tsx" files too? I'm also not a huge fan of naming everything "blah-utils" anymore since the parent folder is already called "utils" and the naming is a bit vague, but that's out of scope here.

sample-app/src/utils/geolocationutils.tsx
yen-tt: I briefly tried separating it out before this but thought it might be a little overkill too since the components would have to manage the async stuff and piece together multiple functions . But yeah I can change so it's not a single method

sample-app/src/utils/geolocationutils.tsx
oshi97: since the search bar executes autocomplete requests, would it be possible to pass along the input intents? Maybe by awaiting the autocomplete request? I had a potentially jank idea where we store the most recent autocomplete request's promise as a ref, and always await it before making searches.

sample-app/src/utils/geolocationutils.tsx
yen-tt: hm yea, will change the file extension. As for the name, tom suggested `SearchHandler` to change this into a class, so I will probably name the file along that line without 'util'

sample-app/src/utils/geolocationutils.tsx
oshi97: It would get us out of making duplicate autocomplete requests before every search, though. I wouldn't be surprised if there's also a nicer way to implement it.

oshi97: We could pass in the inputIntents as a param to this method, so that cases like searching on page load will still fire the autocomplete to check for "near me"

yen-tt: oh for some reason, I was thinking vertical/universal search page wouldn't be able to get intents from search bar for the default search, but I forgot since it's an initial search, we would execute an autocomplete request anyway for the intents. As for searchbar, I think that idea would work. I will something out

sample-app/src/components/SearchBar.tsx
yen-tt: update: split into functions inside `SearchHandler` class as suggested in another comment

sample-app/src/utils/geolocationutils.tsx
yen-tt: update: split into functions inside `SearchHandler` class as suggested in another comment

sample-app/src/utils/geolocationutils.tsx
yen-tt: updated to have it as a class with static functions

sample-app/src/utils/geolocationutils.tsx
yen-tt: updated!

sample-app/src/utils/searchhandler.ts
```diff
@@ -0,0 +1,47 @@
...
+};
+
+/**
+ * Provide utility functions related to vertical and universal search execution.
+ * This includes any potential data check or setup related to a query before
+ * performing a search, such as fetching searchIntents using autocomplete requests,
+ * and retrieving user's location using nagivator.geolocation API.
+ */
+export default class SearchHandler {
+  static async executeSearch(answersActions: AnswersActions, isVertical: boolean) {
```
oshi97, line 17: Are there advantages to putting static methods on a class, as opposed to just having separate methods? Maybe we could use a typescript name space instead? afaik static methods are not treeshakeable

sample-app/src/components/SearchBar.tsx
```diff
@@ -6,28 +6,50 @@ import '../sass/SearchBar.scss';
...
+    executeAutocomplete
+  ] = useAutocomplete(isVertical);
 
-  function executeQuery () {
-    isVertical
-      ? answersActions.executeVerticalQuery()
-      : answersActions.executeUniversalQuery();
+  async function executeQuery () {
+    const responseToLatestRequest = await responseToLatestRequestRef.current;
+    if (responseToLatestRequest?.inputIntents.includes(SearchIntent.NearMe)) {
```
oshi97, line 41: can we extract this method (except for waiting on the autocomplete request on line 40) somewhere, since it's being used in a number of places? I was thinking `executeSearchWIthIntents(intents: SearchIntent[])` and we could move it to `utils/executeSearchWithIntents.ts`?

sample-app/src/hooks/useAutocomplete.tsx
```diff
@@ -1,19 +1,30 @@
...
-import { AutocompleteResult, useAnswersActions } from '@yext/answers-headless-react';
+import { MutableRefObject, useRef, useState } from "react";
+import { AutocompleteResponse, useAnswersActions } from '@yext/answers-headless-react';
 
-export function useAutocomplete(isVertical: boolean): [AutocompleteResult[], () => Promise<void>] {
+export function useAutocomplete(
+  isVertical: boolean
+): [ 
+    AutocompleteResponse|undefined,
+    MutableRefObject<Promise<AutocompleteResponse|undefined>>,
```
oshi97, line 8: Would it make sense to have the searchbar keep track of the latest autocomplete response ref? Instead of the hook, since I'm not sure all users of the hook need this. Alternatively I'd probably move it to the end of the returned tuple, since it's more niche than the other 2

sample-app/src/utils/searchhandler.ts
yen-tt: The class is just a way to group these functions into a unit related to search. With in being a class, I initially consider having a constructor to pass in answersAction and isVertical but then not all functions really not that (getUserLocation). I made them statics since it's not necessary to instantiate new SearchHandler. I don't have a strong preference between utility classes vs utility functions

sample-app/src/components/SearchBar.tsx
yen-tt: sure thing, added to SearchHandler for now, but if we decide to go with separate functions than I will move it out to another file

sample-app/src/hooks/useAutocomplete.tsx
yen-tt: yea that makes sense. I updated for the function returned from the hook to return the promise so that should be cleaner now!

sample-app/src/components/SearchBar.tsx
```diff
@@ -6,28 +6,38 @@ import '../sass/SearchBar.scss';
...
+  placeholder,
+  isVertical,
+  geolocationOptions,
+  screenReaderInstructionsId 
+}: Props) {
   const answersActions = useAnswersActions();
   const query = useAnswersState(state => state.query.input);
-  const [ autocompleteResults, executeAutocomplete ] = useAutocomplete(isVertical);
   const isLoading = useAnswersState(state => state.searchStatus.isLoading);
+  const responseToLatestRequestRef = useRef<Promise<AutocompleteResponse|undefined>>(Promise.resolve(undefined));
```
oshi97, line 34: There should never be a case where a search is executed before an autocomplete request is started (I think?) so the default of Promise.resolve(undefined) should never be needed? I also think that, if such a case were to come up, instead of assuming there are no search intents, we'd want to fire that extra autocomplete request to check for them first.

sample-app/src/utils/searchhandler.ts
```diff
@@ -0,0 +1,68 @@
...
+    isVertical
+      ? answersActions.executeVerticalQuery()
+      : answersActions.executeUniversalQuery();
+    return;
+  }
+
+  static async executeSearchWithIntents(
```
oshi97, line 25: I think it might be a little clearer to have this method separate from the other 3. In my mind, since it uses the other 3, `executeSearchWithIntents` is at a higher level of abstraction than the others. The other 3 are bricks while this one is... whatever you feel like making with bricks

sample-app/src/components/SearchBar.tsx
yen-tt: that's fair. will remove

sample-app/src/utils/searchhandler.ts
yen-tt: Like in a different file? Although it's at a higher level of abstraction, it's still a utility function for search execution so I think it make sense to group them? I imagined it's similar to appliedfilterutils, where there are separate `brick` functions but then there are ones like `getGroupedAppliedFilters` that uses other `brick` functions but overall is still consider a utility function for applied filters.

oshi97: ok sgtm! I liked the organization a lot before but this makes sense too

sample-app/src/utils/search-operations.ts
```diff
@@ -0,0 +1,73 @@
...
+export async function getUserLocation(geolocationOptions?: PositionOptions): Promise<GeolocationPosition> {
+  return new Promise((resolve, reject) => {
+    if ('geolocation' in navigator) {
+      navigator.geolocation.getCurrentPosition(
+        position => resolve(position),
+        err => { 
+          console.error('Error occured using geolocation API. Unable to determine user\'s location.');
+          reject(err); 
+        },
+        Object.assign(defaultGeolocationOptions, geolocationOptions)
```
cea2aj, line 67: It seems odd to me that this function has the side effect of modifying the defaultGeolocationOptions

sample-app/src/components/LocationBias.tsx
```diff
@@ -41,20 +33,17 @@ export default function LocationBias(props: Props) {
...
-      (err) => console.error(err),
-      geolocationOptions);
+  async function handleGeolocationClick() {
+    try {
+      const position = await getUserLocation(geolocationOptions);
+      answersActions.setUserLocation({
+        latitude: position.coords.latitude,
+        longitude: position.coords.longitude,
+      });
+    } catch (e) {
```
cea2aj, line 43: Would it be helpful to log an error about failing to get the user's location?

sample-app/src/components/SearchBar.tsx
```diff
@@ -6,28 +6,38 @@ import '../sass/SearchBar.scss';
...
+import { useRef } from 'react';
+import { AutocompleteResponse } from '@yext/answers-headless';
+import { executeSearchWithIntents } from '../utils/search-operations';
 
 const SCREENREADER_INSTRUCTIONS = 'When autocomplete results are available, use up and down arrows to review and enter to select.'
 
 interface Props {
   placeholder?: string,
   isVertical: boolean,
+  geolocationOptions?: PositionOptions,
```
cea2aj, line 18: do we need to add this to the props, or can we use the default for now?

sample-app/src/pages/VerticalSearchPage.tsx
```diff
@@ -68,7 +69,11 @@ export default function VerticalSearchPage(props: {
       universal: {}
     });
     answersActions.setVerticalKey(props.verticalKey);
-    answersActions.executeVerticalQuery();
+    const executeQuery = async () => {
+      const searchIntents = await getSearchIntents(answersActions, false);
```
cea2aj, line 73: If NEAR_ME stays in the answers state, will this cause searches to always fire an autocomplete request to get the search intents?

sample-app/src/components/SearchBar.tsx
yen-tt: Tom suggested to add this since the SDK's SearchBar allows you to configure the GeoLocation options

sample-app/src/components/LocationBias.tsx
yen-tt: yeah, that error is log in getUserLocation on geolocation API failure, in the case that geolocation API is not accessible, it would reject with a message, where locationbias would conssor error out in line 44

sample-app/src/utils/search-operations.ts
yen-tt: good point, I will update this to `{...defaultGeolocationOptions, ... geolocationOptions}` instead!

sample-app/src/pages/VerticalSearchPage.tsx
yen-tt: The intents store in state may not match with the new query that's going to execute next. So the current stored query may have a near me intent but the newly typed input may not, in which case, an autocomplete request is always executed to ensure the intents we are using match with the upcoming input and not what's already executed. tldr, we can't rely on the near me intent store in state

sample-app/src/components/SearchBar.tsx
```diff
@@ -6,28 +6,38 @@ import '../sass/SearchBar.scss';
...
+  placeholder,
+  isVertical,
+  geolocationOptions,
+  screenReaderInstructionsId 
+}: Props) {
   const answersActions = useAnswersActions();
   const query = useAnswersState(state => state.query.input);
-  const [ autocompleteResults, executeAutocomplete ] = useAutocomplete(isVertical);
   const isLoading = useAnswersState(state => state.searchStatus.isLoading);
+  const responseToLatestRequestRef = useRef<Promise<AutocompleteResponse|undefined>>();
```
cea2aj, line 34: What's the reason for using `useRef` here rather than `useState`?

sample-app/src/components/SearchBar.tsx
```diff
@@ -6,28 +6,38 @@ import '../sass/SearchBar.scss';
...
+  const [ autocompleteResponse, executeAutocomplete] = useAutocomplete(isVertical);
 
-  function executeQuery () {
-    isVertical
-      ? answersActions.executeVerticalQuery()
-      : answersActions.executeUniversalQuery();
+  async function executeQuery () {
+    const responseToLatestRequest = await responseToLatestRequestRef.current;
```
cea2aj, line 38: Would `latestAutocompleteResponse` be clearer?

sample-app/src/components/SearchBar.tsx
yen-tt: we don't want to re-render when this variable changes. It's just a reference so when a search is executed, it can wait for this response to latest request before executing the query, which then would want to trigger a re-render.

cea2aj: Got it. Can you add a comment that indicates that's why a Ref is being used?

sample-app/src/utils/search-operations.ts
```diff
@@ -0,0 +1,73 @@
...
+  const results = isVertical
+    ? await answersActions.executeVerticalAutocomplete()
+    : await answersActions.executeUniversalAutocomplete();
+  return results?.inputIntents;
+}
+
+/**
+ * Retrieves user's location using nagivator.geolocation API
+ */
+export async function getUserLocation(geolocationOptions?: PositionOptions): Promise<GeolocationPosition> {
```
cea2aj, line 58: In the SDK, we don't try to get the location again if the location is already in storage. Should we have that behavior here as well? The API can take a while to get a location, so it could slow things down if we use `navigator.geolocation.getCurrentPostion` on every near me query

sample-app/src/components/SearchBar.tsx
yen-tt: sure thing

sample-app/src/pages/VerticalSearchPage.tsx
cea2aj: Not relying on the near me in state makes sense. It's different than the SDK, however with headless, we can't make the same assumptions

sample-app/src/utils/search-operations.ts
cea2aj: Similarly, if the location is already in storage, we don't need to fire off an additional autocomplete request to check the search intents

sample-app/src/components/SearchBar.tsx
yen-tt: hmm it's not exactly the latest autocomplete response, but rather the promise stored in this ref is tied to the latest request. Let's say there's an autocomplete request #1 trigger before and an autocomplete request #2 trigger after a user hit 'enter' to execute search, and response #2 comes in. #2 would be the latest autocomplete response, but the await here is waiting on the response to the latest request right before the search execution.

sample-app/src/utils/search-operations.ts
yen-tt: I see. I think we should leave `getUserLocation` as is, since user may want to manually update their location through the button in `LocationBias` (which uses this function) if their location did change somehow. But I will update `executeSearchWithIntents` to first check the state before calling this function!

sample-app/src/components/SearchBar.tsx
cea2aj: got it. I was also thinking that it would be helpful to include 'Autcomplete' in the name somewhere, however I'll leave that up to you if we should include that because that would result in a really long name

yen-tt: I think it's ok to keep it as is for now? If there will be more responses to manage in searchBar in the future then we can make that differentiation more distinct in the names

sample-app/src/pages/UniversalSearchPage.tsx
```diff
@@ -17,7 +19,14 @@ export default function UniversalSearchPage(props: { universalResultsConfig: Uni
       vertical: {}
     })
     answersActions.setVerticalKey('');
-    answersActions.executeUniversalQuery();
+    const executeQuery = async () => {
+      let searchIntents: SearchIntent[] = [];
```
cea2aj, line 23: This logic looks correct, however I find it a little difficult to follow since it requires me to go back and forth between the search-operations class to understand it. Would something like this work: `if ( state.location.userLocation ) {
   executeSearch()
} else {
   const intents = getSearchIntents()
   updateLocationIfNeeded(intents)
   executeSearch()
}`

tmeyer2115 (approved):

sample-app/src/pages/UniversalSearchPage.tsx
yen-tt: in both condition, it would still call execute search so I think it's better to just extracted out of that logic and call it once. But I updated the function to be updateLocationIfNeeded as suggested, hopefully that's easier to understand

cea2aj: looks good!

cea2aj (approved):

oshi97 (approved):

PR #60, opened 2021-11-24
Refactor InputDropdown and Dropdown

nmanu1:
Introduces a DropdownSection component that handles the logic of navigating the focus between options within that section of the dropdown. InputDropdown now only handles navigation between which section is focused and it no longer controls how the dropdown is structured.

J=SLAP-1722
TEST=manual

Check that the search bar in the sample-app behaves as expected


sample-app/src/components/DropdownSection.tsx
```diff
@@ -0,0 +1,108 @@
...
+import { useState, useEffect } from "react";
+
+export interface Option {
+  value: string,
+  display: JSX.Element
+}
+
+interface DropdownSectionProps<T> {
+  focusStatus?: string,
+  options: T[],
```
cea2aj, line 11: We may not need a generic for the options if each option has a display prop

sample-app/src/components/DropdownSection.tsx
```diff
@@ -0,0 +1,108 @@
...
+  onFocusChange,
+  onLeaveSectionFocus = () => {},
+  onClickOption,
+  label = '',
+  cssClasses
+}: DropdownSectionProps<T>): JSX.Element | null {
+
+  const [focusedOptionIndex, setFocusedOptionIndex] = useState<number>(0);
+
+  function changeOptionFocus(focusNext: boolean) {
```
cea2aj, line 39: What do you think of having `incrementOptionFocus` and `decrementOptionFocus` instead? I think it's a little clearer than the `focusNext` boolean flag

sample-app/src/components/DropdownSection.tsx
```diff
@@ -0,0 +1,108 @@
...
+  value: string,
+  display: JSX.Element
+}
+
+interface DropdownSectionProps<T> {
+  focusStatus?: string,
+  options: T[],
+  optionIdPrefix: string,
+  onFocusChange: (value: string, focusedOptionId: string) => void,
+  onLeaveSectionFocus?: (focusNext: boolean) => void,
```
cea2aj, line 14: What do you think of `pastEnd` instead of `focusNext`? To me, `focusNext` implies that the focus should move to the next section, however there might not be another section (e.g. if it's the last section)

sample-app/src/components/InputDropdown.tsx
```diff
@@ -52,74 +47,120 @@ export default function InputDropdown({
...
     setScreenReaderKey(0);
   }
 
+  const childrenArray = React.Children.toArray(children);
+  const childrenWithProps = childrenArray.map((child, index) => {
+    if (!React.isValidElement(child)) {
+      return child;
+    }
+    
+    const modifiedOnClickOption : typeof child.props.onClickOption = (option: typeof child.props.options[0]) => {
```
cea2aj, line 83: is this `typeof` working? I'd be surprised if it works if the props type is `any`

sample-app/src/components/InputDropdown.tsx
```diff
@@ -52,74 +47,120 @@ export default function InputDropdown({
...
+
+    const modifiedOnFocusChange = (value: string, focusedOptionId: string) => {
+      child.props.onFocusChange(value);
+      setFocusedOptionId(focusedOptionId);
+    };
+
+    if (focusedSectionIndex === undefined) {
+      return React.cloneElement(child, { onLeaveSectionFocus, focusStatus: 'reset', onClickOption: modifiedOnClickOption });
+    } else if (index === focusedSectionIndex) {
+      return React.cloneElement(child, { onLeaveSectionFocus, focusStatus: 'active', onFocusChange: modifiedOnFocusChange, onClickOption: modifiedOnClickOption });
```
cea2aj, line 96: nit: we may want some new lines depending on how long this line is

sample-app/src/components/InputDropdown.tsx
```diff
@@ -52,74 +47,120 @@ export default function InputDropdown({
...
+      let newSectionIndex: number | undefined = focusNext ? focusedSectionIndex + 1 : focusedSectionIndex - 1;
+      if (newSectionIndex < 0) {
+        newSectionIndex = undefined;
+        onInputChange(latestUserInput);
+      }
+      dispatch({ type: 'FocusSection', newIndex: newSectionIndex });
+    } else if (focusedSectionIndex === numSections - 1 && !focusNext) {
+      let newSectionIndex: number | undefined = focusedSectionIndex - 1;
+      if (newSectionIndex < 0) {
+        newSectionIndex = undefined;
```
cea2aj, line 117: should we also call `onInputChange` here?

sample-app/src/components/InputDropdown.tsx
```diff
@@ -52,74 +47,120 @@ export default function InputDropdown({
...
-        : latestUserInput;
-      updateInputValue(newValue);
+      dispatch({ type: 'HideSections' });
+    } else if ((evt.key === 'ArrowDown' || evt.key === 'ArrowRight' ) && numSections > 0 && focusedSectionIndex === undefined) {
+      dispatch({ type: 'FocusSection', newIndex: 0 });
     }
   }
 
+  console.log(numSections);
+  console.log(focusedSectionIndex);
```
cea2aj, line 162: console.log

sample-app/src/components/InputDropdown.tsx
```diff
@@ -128,52 +169,40 @@ export default function InputDropdown({
           placeholder={placeholder}
           onChange={evt => {
             const value = evt.target.value;
-            dispatch({ type: 'ShowOptions' });
+            dispatch({ type: 'ShowSections' });
             setLatestUserInput(value);
-            updateInputValue(value);
-            updateDropdown();
+            onInputChange(value);
```
cea2aj, line 174: I feel like `setInputValue` is a more accurate name. Do you think that works better than `onInputChange`?

sample-app/src/components/SearchBar.tsx
```diff
@@ -68,29 +83,43 @@ export default function SearchBar({
...
-          return {
-            value: result.value,
-            render: () => renderWithHighlighting(result)
-          }
-        }) ?? []}
-        optionIdPrefix='Autocomplete__option'
+        screenReaderText={screenReaderText}
         onSubmit={executeQuery}
-        updateInputValue={value => {
+        onInputChange={(value) => {
```
cea2aj, line 88: nit: we usually don't put the parenthesis around the param when there is just one param

sample-app/src/components/SearchBar.tsx
```diff
@@ -68,29 +83,43 @@ export default function SearchBar({
...
           inputElement: 'SearchBar__input',
           inputContainer: 'SearchBar__inputContainer'
         }}
-      />
+      >
+        {
+          options.length > 0 &&
+          <DropdownSection<Option>
+            options={options}
+            optionIdPrefix={`Autocomplete__option-${0}`}
```
cea2aj, line 105: Why is the zero wrapped in curly braces?

sample-app/src/components/DropdownSection.tsx
```diff
@@ -0,0 +1,108 @@
...
+  options,
+  optionIdPrefix,
+  onFocusChange,
+  onLeaveSectionFocus = () => {},
+  onClickOption,
+  label = '',
+  cssClasses
+}: DropdownSectionProps<T>): JSX.Element | null {
+
+  const [focusedOptionIndex, setFocusedOptionIndex] = useState<number>(0);
```
yen-tt, line 37: should the initial state be -1 instead? since section is not focused by default.

sample-app/src/components/DropdownSection.tsx
yen-tt: agree with splitting into increment/decrement , I think `focusNext` boolean can be a bit confusing

sample-app/src/components/DropdownSection.tsx
```diff
@@ -0,0 +1,108 @@
...
+      } else if (evt.key === 'ArrowUp' || evt.key === 'ArrowLeft') {
+        changeOptionFocus(false);
+      }
+    }
+  }
+
+  useEffect(() => {
+    if (focusStatus === 'active') {
+      onFocusChange(options[focusedOptionIndex].value, `${optionIdPrefix}-${focusedOptionIndex}`);
+    } else if (focusStatus === 'reset') {
```
yen-tt, line 69: can we get rid of 'reset' status? from what I can see, it's use when focusedSectionIndex is undefined, which is when everything is hidden anyway. Not much different from 'inactive' status

sample-app/src/components/DropdownSection.tsx
```diff
@@ -0,0 +1,108 @@
...
+      setFocusedOptionIndex(0);
+    }
+  }, [focusStatus]);
+
+  useEffect(() => {
```
yen-tt, line 74: can this useEffect be group with the useEffect above?

sample-app/src/components/InputDropdown.tsx
```diff
@@ -52,74 +47,120 @@ export default function InputDropdown({
...
   const inputRef = useRef<HTMLInputElement>(document.createElement('input'));
+  const dropdownRef = useRef<HTMLDivElement>(document.createElement('div'));
 
   if (!shouldDisplayDropdown && screenReaderKey) {
     setScreenReaderKey(0);
   }
 
+  const childrenArray = React.Children.toArray(children);
+  const childrenWithProps = childrenArray.map((child, index) => {
+    if (!React.isValidElement(child)) {
```
yen-tt, line 79: Should probably check if this child is a Section component. Otherwise child.props.onClickOption below would be undefined and would throw an error when called

sample-app/src/components/InputDropdown.tsx
```diff
@@ -52,74 +47,120 @@ export default function InputDropdown({
...
+      return child;
+    }
+    
+    const modifiedOnClickOption : typeof child.props.onClickOption = (option: typeof child.props.options[0]) => {
+      child.props.onClickOption(option);
```
yen-tt, line 84: Should probably check if this child is a Section component. Otherwise child.props.onClickOption would be undefined and would throw an error when call

sample-app/src/components/InputDropdown.tsx
```diff
@@ -52,74 +47,120 @@ export default function InputDropdown({
...
+    } else if (index === focusedSectionIndex) {
+      return React.cloneElement(child, { onLeaveSectionFocus, focusStatus: 'active', onFocusChange: modifiedOnFocusChange, onClickOption: modifiedOnClickOption });
+    } else {
+      return React.cloneElement(child, { onLeaveSectionFocus, focusStatus: 'inactive', onClickOption: modifiedOnClickOption });
+    }
+  });
+
+  const numSections = childrenWithProps.length;
+
+  function onLeaveSectionFocus(focusNext: boolean) {
```
yen-tt, line 104: is there a more descriptive name than 'focusNext'?

sample-app/src/components/InputDropdown.tsx
```diff
@@ -52,74 +47,120 @@ export default function InputDropdown({
...
+    if (focusedSectionIndex === undefined && focusNext) {
+      dispatch({ type: 'FocusSection', newIndex: 0 });
+    } else if (focusedSectionIndex !== undefined && focusedSectionIndex < numSections - 1) {
+      let newSectionIndex: number | undefined = focusNext ? focusedSectionIndex + 1 : focusedSectionIndex - 1;
```
yen-tt, line 108: nit: new line on ? and :

sample-app/src/components/InputDropdown.tsx
```diff
@@ -52,74 +47,120 @@ export default function InputDropdown({
...
+      return React.cloneElement(child, { onLeaveSectionFocus, focusStatus: 'inactive', onClickOption: modifiedOnClickOption });
+    }
+  });
+
+  const numSections = childrenWithProps.length;
+
+  function onLeaveSectionFocus(focusNext: boolean) {
+    if (focusedSectionIndex === undefined && focusNext) {
+      dispatch({ type: 'FocusSection', newIndex: 0 });
+    } else if (focusedSectionIndex !== undefined && focusedSectionIndex < numSections - 1) {
```
yen-tt, line 107: I think focusedSectionIndex !== undefined can be removed

sample-app/src/components/InputDropdown.tsx
```diff
@@ -52,74 +47,120 @@ export default function InputDropdown({
...
+      dispatch({ type: 'FocusSection', newIndex: newSectionIndex });
+    } else if (focusedSectionIndex === numSections - 1 && !focusNext) {
+      let newSectionIndex: number | undefined = focusedSectionIndex - 1;
+      if (newSectionIndex < 0) {
+        newSectionIndex = undefined;
+      }
+      dispatch({ type: 'FocusSection', newIndex: newSectionIndex });
+    }
+  }
+
```
yen-tt, line 122: This seems a bit difficult to follow, can we refactor this top level else if to something like this?
```
let moveToNextSection = focusNext and NOT last section;
let movePreviousSection = !focusNext and NOT first section;
else if it's moveToNextSection OR movePreviousSection
	update input + move to new section
else if it's (!focusNext and first section) //want to move up to search bar
	update to Latest input + section index as undefined
```

sample-app/src/components/InputDropdown.tsx
```diff
@@ -52,74 +47,120 @@ export default function InputDropdown({
...
   function handleDocumentClick(evt: MouseEvent) {
     const target = evt.target as HTMLElement;
-    if (!target.isSameNode(inputRef.current)) {
-      dispatch({ type: 'HideOptions' });
+    if (!(target.isSameNode(inputRef.current) || (dropdownRef.current !== null && dropdownRef.current.contains(target)))) {
```
yen-tt, line 125: would optional changing work here? so just dropdownRef.current?.contains(target) without the null check

sample-app/src/components/InputDropdown.tsx
nmanu1: no, it's not working here, so I can remove it and address the typing later with the type guarding you mentioned. but, if we want to avoid having `(option: any)` in this PR, then I could also do the type guarding now

sample-app/src/components/InputDropdown.tsx
cea2aj: I think Yen means "optional chaining"

sample-app/src/components/DropdownSection.tsx
nmanu1: right now the logic works by always pointing to the first valid option index (which would be 0). that way, as soon as focus shifts to it, the index will be valid. `focusStatus` is indicating whether the section is in focus or not

sample-app/src/components/DropdownSection.tsx
nmanu1: the 'reset' status is how the sections know to reset their `focusedOptionIndex` to the beginning of the section. otherwise, it could result in incorrect option focus behavior after a search is performed. because the section would remember the last place it was focused as the default

sample-app/src/components/InputDropdown.tsx
```diff
@@ -52,74 +47,120 @@ export default function InputDropdown({
...
+    }
+  }
+
+  function handleDocumentKeydown(evt: globalThis.KeyboardEvent) {
+    if ((evt.key === 'Escape' || evt.key === 'Tab')) {
+      dispatch({ type: 'HideSections' });
     }
   }
 
   useEffect(() => {
```
yen-tt, line 136: Should this get call every render? or just 'on mount' or on first render. If it is not on every render, I think we should add '[]' as the second argument for this useEffect

sample-app/src/components/DropdownSection.tsx
nmanu1: the event listener needs to be added whenever the section is active, not just when it first becomes active (like the first useEffect). I tried combining them, but then only the first keydown event is registered

sample-app/src/components/InputDropdown.tsx
```diff
@@ -52,74 +47,120 @@ export default function InputDropdown({
...
+      document.addEventListener('click', handleDocumentClick);
+      document.addEventListener('keydown', handleDocumentKeydown);
+      return () => {
+        document.removeEventListener('click', handleDocumentClick);
+        document.removeEventListener('keydown', handleDocumentKeydown);
+      }
+    }
   });
 
   function onKeyDown(evt: KeyboardEvent<HTMLInputElement>) {
```
yen-tt, line 147: Is this onKeyDown still needed? Can we remove this and move the Enter functionality to `handleDocumentKeydown` instead?

sample-app/src/components/InputDropdown.tsx
nmanu1: yes, this needs to be called every render, because otherwise it would be removed after the first render and then never added again

sample-app/src/components/InputDropdown.tsx
nmanu1: `onKeyDown` tracks the keys pressed when the user has clicked into the input bar. otherwise, even if the user presses Enter for something else on the page, it would trigger the onSubmit for the input bar

sample-app/src/components/SearchBar.tsx
nmanu1: Oops that's leftover from my testing

sample-app/src/components/InputDropdown.tsx
nmanu1: I think the idea was that this could be used for more than just setting the input. The user can do whatever they want when the input changes, not just set the value

sample-app/src/components/InputDropdown.tsx
nmanu1: yes, that would work!

sample-app/src/components/InputDropdown.tsx
nmanu1: yes, good catch!

sample-app/src/components/DropdownSection.tsx
nmanu1: Oh that's true, because it would be unlikely that any more information is needed for the Option other than the display and what value should be used for the input bar. I'll remove it

sample-app/src/components/DropdownSection.tsx
yen-tt: I think during focus shift, once a section is marked with a focus of 'inactive' then it can reset to the beginning of the section then.

sample-app/src/components/DropdownSection.tsx
yen-tt: `focusedSectionIndex` is undefined by default, in a similar sense I would expect this to be -1 by default as well? From my understanding, all section would be mark as 'inactive' first before setting to 'active', so it can set the index to 0 then (this would help get rid of reset - from my other comment in the pr).

sample-app/src/components/DropdownSection.tsx
yen-tt: you are cloning the element each time when section is set active right? So wouldn't it be first render of the newly clone section be enough to setup the listener?

sample-app/src/components/InputDropdown.tsx
yen-tt: I see, we can keep the enter in onKeyDown then. But what about the arrowDown/Right in the else if? I assume the other `handleKeyDown` listeners should already take care of them no?

sample-app/src/components/InputDropdown.tsx
```diff
@@ -52,74 +47,120 @@ export default function InputDropdown({
...
     const target = evt.target as HTMLElement;
-    if (!target.isSameNode(inputRef.current)) {
-      dispatch({ type: 'HideOptions' });
+    if (!(target.isSameNode(inputRef.current) || (dropdownRef.current !== null && dropdownRef.current.contains(target)))) {
+      dispatch({ type: 'HideSections' });
+    }
+  }
+
+  function handleDocumentKeydown(evt: globalThis.KeyboardEvent) {
+    if ((evt.key === 'Escape' || evt.key === 'Tab')) {
```
yen-tt, line 131: remove extra parentheses

sample-app/src/components/DropdownSection.tsx
nmanu1: if I split it into 2 functions, it would look something like:
```
  function incrementOptionFocus() {
    let newIndex = focusedOptionIndex + 1;
    if (newIndex < options.length) {
      onFocusChange(options[newIndex].value, `${optionIdPrefix}-${newIndex}`);
    } else {
      onLeaveSectionFocus(true);
      newIndex = options.length - 1;
    }
    setFocusedOptionIndex(newIndex);
  }

  function decrementOptionFocus() {
    let newIndex = focusedOptionIndex - 1;
    if (newIndex > -1) {
      onFocusChange(options[newIndex].value, `${optionIdPrefix}-${newIndex}`);
    } else {
      onLeaveSectionFocus(false);
      newIndex = 0;
    }
    setFocusedOptionIndex(newIndex);
  }
```
it seems a little repetitive to me, but if it's more clear, then I can change it to this

cea2aj: I think the increased clarity is worth the repetition (2 lines of code)

sample-app/src/components/DropdownSection.tsx
nmanu1: right now, the way a section is maintaining knowledge of which side it gains focus from (start or end) depends on what the focused option index is when 'inactive'. if this were reset, when a section becomes active, it wouldn't know if it should focus on the last option or the first

sample-app/src/components/InputDropdown.tsx
cea2aj: got it, we can stick with `onInputChange` then

sample-app/src/components/DropdownSection.tsx
nmanu1: ok, changed!

sample-app/src/components/InputDropdown.tsx
nmanu1: added a check

sample-app/src/components/InputDropdown.tsx
nmanu1: I restructured this a little. Does it make more sense now?

sample-app/src/components/InputDropdown.tsx
```diff
@@ -104,32 +107,36 @@ export default function InputDropdown({
...
+        newSectionIndex = numSections - 1;
       }
       dispatch({ type: 'FocusSection', newIndex: newSectionIndex });
     }
   }
 
   function handleDocumentClick(evt: MouseEvent) {
     const target = evt.target as HTMLElement;
-    if (!(target.isSameNode(inputRef.current) || (dropdownRef.current !== null && dropdownRef.current.contains(target)))) {
+    if (!(target.isSameNode(inputRef.current) || (dropdownRef.current?.contains(target)))) {
```
cea2aj, line 126: nit: extra spacing after the curly brackets

sample-app/src/components/InputDropdown.tsx
nmanu1: I restructured it, so now I think the `focusedSectionIndex !== undefined` should stay

sample-app/src/components/InputDropdown.tsx
nmanu1: yes, I moved that into `handleKeyDown`

sample-app/src/components/InputDropdown.tsx
nmanu1: updated!

sample-app/src/components/InputDropdown.tsx
```diff
@@ -52,71 +48,114 @@ export default function InputDropdown({
...
   function handleDocumentClick(evt: MouseEvent) {
     const target = evt.target as HTMLElement;
-    if (!target.isSameNode(inputRef.current)) {
-      dispatch({ type: 'HideOptions' });
+    if (!(target.isSameNode(inputRef.current) || (dropdownRef.current?.contains(target)))) {
+      dispatch({ type: 'HideSections' });
+    }
+  }
+
+  function handleDocumentKeydown(evt: globalThis.KeyboardEvent) {
```
cea2aj, line 131: I think having both `handleDocumentKeydown` and `onKeyDown` might be a little confusing. Do you think it makes sense to rename `onKeyDown` to `handleInputElementKeydown`?

sample-app/src/components/DropdownSection.tsx
nmanu1: It would work either way. I could make it -1 or undefined and then change it to 0 when the section becomes active. In InputDropdown, it makes sense to start with undefined because it has knowledge of the input bar that the section doesn't have. So, I think I slightly prefer it this way so that the sections always point to a valid option index and don't have the notion of pointing to something before their first option.

sample-app/src/components/DropdownSection.tsx
nmanu1: I think that makes sense. I'll change it

sample-app/src/components/InputDropdown.tsx
nmanu1: Yes, that would be more clear about which keydown is being handled

sample-app/src/components/InputDropdown.tsx
cea2aj: Maybe `pastSectionEnd`?

sample-app/src/components/InputDropdown.tsx
nmanu1: Updated!

sample-app/src/components/InputDropdown.tsx
nmanu1: updated to 'pastEnd'

sample-app/src/components/DropdownSection.tsx
nmanu1: updated!

cea2aj (approved):
Looks good!

sample-app/src/components/InputDropdown.tsx
nmanu1: actually, `pastSectionEnd` would be better. but, would that be clear that the opposite would be past the beginning of the section?

sample-app/src/components/DropdownSection.tsx
nmanu1: updated to remove 'reset' status by changing the key of the child component instead

cea2aj (approved):

PR #62, opened 2021-11-29
small nits

oshi97:
* import from answers-headless-react only in the sample app
* eslint rule to remove unnecessary curly braces in props
* add missing useEffect deps

J=None
TEST=manual

### DIFF
--- a/src/AnswersHeadlessContext.ts
+++ b/src/AnswersHeadlessContext.ts
@@ -1,4 +1,4 @@
-import { AnswersHeadless } from '@yext/answers-headless';
+import { AnswersHeadless } from '@yext/answers-headless-react';
 import { createContext } from 'react';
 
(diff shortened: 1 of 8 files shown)


src/AnswersHeadlessContext.ts
```diff
@@ -1,4 +1,4 @@
-import { AnswersHeadless } from '@yext/answers-headless';
+import { AnswersHeadless } from '@yext/answers-headless-react';
```
yen-tt, line 1: should these imports from the top level src folder be importing from its own package? or can it stay as '@yext/answers-headless'

oshi97: my bad I fixed these locally but didn't add them before comitting

yen-tt (approved):

PR #63, opened 2021-11-29
Add Tailwind

cea2aj:
Add Tailwind to the sample app

If we uneject the sample app, we'll need to switch to something like Craco to add Tailwind (see https://tailwindcss.com/docs/guides/create-react-app). However since the app is currently ejected, this will do for now.

J=SLAP-1639
TEST=manual

Add some tailwind styling to the search bar and nav bar (from a future PR) and confirm the tailwind is affecting the styling of the page


sample-app/package.json
```diff
@@ -173,5 +187,10 @@
     "presets": [
       "react-app"
     ]
+  },
+  "devDependencies": {
+    "autoprefixer": "^9.8.8",
+    "postcss": "^7.0.39",
+    "tailwindcss": "npm:@tailwindcss/postcss7-compat@^2.2.17"
```
oshi97, line 194: do we need the npm prefix? I've never seen this one before

cea2aj: I tried removing it, but the site didn't complile. I think it's needed because I'm using the compatability build rather than the the tailwindcss repo itself, and it needs to know to look at the other repo. I was seeing an error with the regular tailwind repo, and it said it's likely to happen if a toolchain hasn't been updated to use postcss 8, and based on my testing, it didn't look like our version of create-react-app worked with it. (See https://tailwindcss.com/docs/installation#post-css-7-compatibility-build)

oshi97: is there anything wrong with using postcss 8?

cea2aj: The version of create-react-app that we ejected at doesn't support it. We would need to bump several dependencies including 'css-loader', 'webpack', 'terser-webpack-plugin' and potentially more, and I'd rather not go down that rabbit hole since long term we are looking to un-eject the app. Potentially a newer version of create react app supports it, however I wasn't able to determine that by quickly looking at the Github repo or the release notes

oshi97: I can't read

oshi97 (approved):

PR #64, opened 2021-11-30
Add FilterSearch component

nmanu1:
Create a component for FilterSearch using the DropdownSection component and the refactored InputDropdown. Generalize the previous `useAutocomplete` hook so it can be used for FilterSearch too. 

J=SLAP-1643
TEST=manual

Conduct filter searches in sample-app to check the behavior is as expected. Check that search bar functions as before.


sample-app/src/components/FilterSearch.tsx
```diff
@@ -0,0 +1,125 @@
...
+        phrase: `${optionInfo} autocomplete option found.`,
+        pluralForm: `${optionInfo} autocomplete options found.`,
+        count: section.results.length
+      });
+    });
+    screenReaderText = screenReaderPhrases.join(' ');
+  }
+
+  return (
+    <div className='FilterSearch'>
```
yen-tt, line 71: I think we want cssClasses specified in the props now, instead of hardcoding them in, like what you have for InputDowndown!

sample-app/src/components/InputDropdown.tsx
```diff
@@ -157,8 +160,8 @@ export default function InputDropdown({
   });
 
   function handleInputElementKeydown(evt: KeyboardEvent<HTMLInputElement>) {
-    if (evt.key === 'Enter') {
-      onInputChange(inputValue);
+    if (evt.key === 'Enter' && focusedSectionIndex === undefined && !onlySubmitOnOption) {
```
yen-tt, line 163: is `focusedSectionIndex === undefined` needed? this is only trigger by onKeyDown in input element so would `focusedSectionIndex` ever be defined if we are in input element?

sample-app/src/pages/VerticalSearchPage.tsx
```diff
@@ -11,7 +11,8 @@ import '../sass/VerticalSearchPage.scss';
 import { StandardCard } from '../components/cards/StandardCard';
 import { useLayoutEffect } from 'react';
 import { useAnswersActions } from '@yext/answers-headless-react';
-import { SearchIntent } from '@yext/answers-headless-react';
+import FilterSearch from '../components/FilterSearch';
+import { SearchIntent } from '@yext/answers-headless';
```
yen-tt, line 15: let's keep the imports to be from `@yext/answers-headless-react`

sample-app/src/components/DropdownSection.tsx
```diff
@@ -59,24 +59,24 @@ export default function DropdownSection({
...
+      onSubmit(focusedOptionIndex);
     }
   }
 
   useEffect(() => {
     if (focusStatus === 'active') {
       onFocusChange(options[focusedOptionIndex].value, `${optionIdPrefix}-${focusedOptionIndex}`);
     }
-  }, [focusStatus, focusedOptionIndex, onFocusChange, optionIdPrefix, options]);
+  }, [focusStatus]);
```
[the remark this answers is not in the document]
nmanu1: Yes, having the additional dependencies was causing the `onFocusChange` to be called more often, which caused the wrong text to be shown in the input bar after navigating back to it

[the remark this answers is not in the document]
yen-tt: can you try using `useCallback` hook for those additional dependencies and called that memorized function instead in useEffect.

sample-app/src/components/FilterSearch.tsx
nmanu1: Oh ok, so all the cssClasses that we pass in for InputDropdown and DropdownSection should be specified in props for FilterSearch? And similarly for SearchBar?

yen-tt: double check this with connor, but yea I believe the css classes should be specified in the parent component

sample-app/src/pages/VerticalSearchPage.tsx
nmanu1: fixed

sample-app/src/components/FilterSearch.tsx
```diff
@@ -0,0 +1,125 @@
...
+  searchFields: Omit<SearchParameterField, 'fetchEntities'>[],
+  screenReaderInstructionsId: string
+}
+
+export default function FilterSearch (props: FilterSearchProps): JSX.Element {
+  const { title, sectioned, searchFields, screenReaderInstructionsId } = props;
+  const answersActions = useAnswersActions();
+  const [input, setInput] = useState('');
+  const searchParamFields = searchFields.map((searchField) => {
+    return { ...searchField, fetchEntities: false }
```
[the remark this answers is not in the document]
cea2aj: It's something that the API supports, but our filter search component in the SDK doesn't use it

sample-app/src/components/FilterSearch.tsx
```diff
@@ -0,0 +1,125 @@
...
+  });
+
+  const [filterSearchResponse, executeFilterSearch] = useSynchronizedSearch((inputValue) => {
+    return inputValue
```
cea2aj, line 29: Rather than using the ternary operator here, can we do something like: `return answersActions.executeFilterSearch(inputValue ?? '', sectioned, searchParamFields);`

sample-app/src/components/DropdownSection.tsx
```diff
@@ -94,7 +94,7 @@ export default function DropdownSection({
         key={index}
         className={className}
         id={`${optionIdPrefix}-${index}`}
-        onClick={() => onClickOption(option, index)}>
+        onClick={() => onSubmit(index, option)}>
```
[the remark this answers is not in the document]
nmanu1: Yeah, that makes sense. Since the option is optional for this function while the index is required, I get an error if I have the param ordering switched. So, I can switch the ordering for renderOption to have it be consistent

sample-app/src/components/FilterSearch.tsx
nmanu1: Updated!

sample-app/src/components/DropdownSection.tsx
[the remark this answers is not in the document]
nmanu1: Done!

sample-app/src/components/FilterSearch.tsx
cea2aj: Yes, we are going to have all the css classes available in the props. No need to worry about it for the search bar since I will update that in my PR

sample-app/src/components/InputDropdown.tsx
```diff
@@ -49,6 +50,7 @@ export default function InputDropdown({
   screenReaderInstructions,
   screenReaderInstructionsId,
   screenReaderText,
+  onlySubmitOnOption,
```
cea2aj, line 53: Can you explain the purpose of this option?

nmanu1: It controls whether you want to allow users to submit only when they've selected a dropdown option, or if they can submit whatever they've typed.

In the case of the SearchBar, they should be allowed to submit anything, but for FilterSearch, they can only select a filter from the dropdown and nothing should happen if they press Enter on their own text without focusing on a dropdown option.

cea2aj: Got it. What do you think of the name`onlyAllowDropdownOptionSubmissions`instead?

sample-app/src/components/DropdownSection.tsx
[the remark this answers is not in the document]
cea2aj: Can you explain why index is required and option is optional?

sample-app/src/components/DropdownSection.tsx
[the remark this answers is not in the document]
oshi97: I think generally speaking you shouldn't rely on useCallback and useMemo to ensure correct behavior. This is the original tweet from Dan Abramov https://twitter.com/dan_abramov/status/1085685515951714306?lang=en and hooks faq part https://reactjs.org/docs/hooks-faq.html#how-to-memoize-calculations
You also don't want to "lie" to react about dependencies, which is a really common hack people use but can lead to hard to find bugs. Do you know exactly how the deps are causing extra calls? There's normally a way to write it without hiding deps

sample-app/src/components/InputDropdown.tsx
```diff
@@ -174,13 +177,13 @@ export default function InputDropdown({
...
             onInputChange(value);
-            onInputFocus();
+            onInputFocus(value);
             setChildrenKey(childrenKey + 1);
             dispatch({ type: 'ShowSections' });
             setScreenReaderKey(screenReaderKey + 1);
           }}
           onClick={() => {
-            onInputFocus();
+            onInputFocus(inputValue);
```
oshi97, line 186: what is the difference between latestUserInput and inputValue?

sample-app/src/hooks/useSynchronizedSearch.tsx
```diff
@@ -0,0 +1,24 @@
+import { useRef, useState } from "react";
+
+export function useSynchronizedSearch<T>(
```
oshi97, line 3: maybe `useSyncronizedRequest`? nothing here is specific to searching

sample-app/src/components/InputDropdown.tsx
```diff
@@ -157,8 +160,8 @@ export default function InputDropdown({
   });
 
   function handleInputElementKeydown(evt: KeyboardEvent<HTMLInputElement>) {
-    if (evt.key === 'Enter') {
-      onInputChange(inputValue);
+    if (evt.key === 'Enter' && focusedSectionIndex === undefined && !onlySubmitOnOption) {
+      setLatestUserInput(inputValue);
       onSubmit(inputValue);
```
oshi97, line 165: instead of adding a new `onlySubmitOnOption` prop, would it be simpler/more flexible to pass `focusedSectionIndex` as a param to `onSubmit`?

sample-app/src/components/DropdownSection.tsx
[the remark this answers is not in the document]
nmanu1: I could add a `useRef` variable that tracks if the dropdown section was active on the last render or not. That way I could call `onFocusChange` only when the `focusStatus` changes from inactive to active without having any dependencies.

sample-app/src/components/DropdownSection.tsx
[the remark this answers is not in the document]
nmanu1: When an option is submitted, FilterSearch would always need to know what the index of the option was to determine which filter to select. But, the option is only needed if we need to access its value to set the text in the input bar. When navigating to the option using arrow keys and selecting with Enter, the text in the input bar has already been set to the correct value and does not need to be set again. However, if an option is clicked, the text in the input bar needs to be set. By making option optional, we can avoid an extra call to set the input text.

But, if it's confusing to have it optional, we could make them both required and just always set the input text on submission, even if it isn't needed for submitting with Enter and would result in an extra render.

sample-app/src/components/InputDropdown.tsx
nmanu1: `inputValue` is the text in the input bar. `latestUserInput` is the last thing typed or queried by the user. the `inputValue` will change as the user navigates and focuses on different options, while `latestUserInput` will stay the same. `latestUserInput` is needed to reset the text in the input bar when the user navigates back to the input bar and no longer focuses on any option

sample-app/src/components/InputDropdown.tsx
nmanu1: That could be used to prevent FilterSearch from submitting, but `setLatestUserInput` would still be called and the dropdown would be hidden when it shouldn't be. According to Rose, it should be apparent that nothing happens if the user presses Enter without selecting/focusing on a dropdown option

sample-app/src/components/DropdownSection.tsx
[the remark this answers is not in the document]
nmanu1: Updated to have `useRef` instead of the `useEffect` with dependencies

sample-app/src/components/FilterSearch.tsx
nmanu1: Oops, didn't see the edited comment. I went ahead and changed it in FilterSearch

sample-app/src/components/InputDropdown.tsx
nmanu1: That's more clear, so I changed it to that

sample-app/src/hooks/useSynchronizedSearch.tsx
nmanu1: Updated search to request and simplified some naming

sample-app/src/hooks/useSynchronizedRequest.tsx
```diff
@@ -0,0 +1,24 @@
...
+export function useSynchronizedRequest<T>(
+  executeRequest: (inputValue?: string) => Promise<T | undefined>
+): [
+    T | undefined,
+    (inputValue?: string) => Promise<T | undefined>
+  ]
+{
+  const networkIds = useRef({ latestRequest: 0, responseInState: 0 });
+  const [synchronizedResponse, setSynchronizedResponse] = useState<T>();
+  async function executeSynchronizedRequest (inputValue?: string): Promise<T | undefined> {
```
yen-tt, line 12: if this is now handling generic requests, lets also make the param generic as well. like ..args? or an object so it can pass in whatever field to executeRequest to use

sample-app/src/components/DropdownSection.tsx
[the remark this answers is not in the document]
cea2aj: I think that we should design the `DropdownSection` in a way that makes the most sense for itself. I would prefer to alway include the option and we can have the `FilterSearch` component be responsible for preventing extra renders if needed. For the scope of this item, I think an extra render is okay.

sample-app/src/components/DropdownSection.tsx
```diff
@@ -12,7 +12,7 @@ export interface DropdownSectionProps {
   optionIdPrefix: string,
   onFocusChange?: (value: string, focusedOptionId: string) => void,
   onLeaveSectionFocus?: (pastSectionEnd: boolean) => void,
-  onClickOption?: (option: Option, optionIndex: number) => void,
+  onSubmit?: (optionIndex: number, option?: Option) => void,
```
cea2aj, line 15: Do you think `onSelect` might make more sense within the scope of a `DropdownSection`? I can see a 'submission' making sense for a search bar, but for a dropdown, I think 'selection' makes more sense

sample-app/src/components/DropdownSection.tsx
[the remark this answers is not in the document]
cea2aj: By making `option` required, we can swap `option` and `index`. I see option as the primary piece of data, and index as more of metadata, so putting option first makes more sense. Also, I'm not sure we need the render function of the option in the callback, so maybe we could use the options's `value` instead?

sample-app/src/components/FilterSearch.tsx
cea2aj: No worries, it will just make our lives easier when we add the tailwind styling

sample-app/src/hooks/useSynchronizedRequest.tsx
nmanu1: updated to take in a generic param argument

sample-app/src/components/DropdownSection.tsx
nmanu1: I wasn't sure if 'select' might be confused with focusing on an option. But maybe since there is an 'onFocusChange', it will be clear

sample-app/src/components/DropdownSection.tsx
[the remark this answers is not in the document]
nmanu1: That makes sense. I'll change it to pass `(optionValue, index)`

sample-app/src/components/DropdownSection.tsx
nmanu1: Updated

sample-app/src/components/DropdownSection.tsx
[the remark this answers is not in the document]
nmanu1: Updated to `onSelect(optionValue, optionIndex)`

sample-app/src/hooks/useSynchronizedRequest.tsx
```diff
@@ -1,18 +1,18 @@
...
 
-export function useSynchronizedRequest<T>(
-  executeRequest: (inputValue?: string) => Promise<T | undefined>
+export function useSynchronizedRequest<RequestParams, ResponseType>(
+  executeRequest: (data?: RequestParams) => Promise<ResponseType | undefined>
 ): [
-    T | undefined,
-    (inputValue?: string) => Promise<T | undefined>
+    ResponseType | undefined,
+    (data?: RequestParams) => Promise<ResponseType | undefined>
```
cea2aj, line 7: Do you think `RequestDataType` or `RequestParamType` might be more accurate? `RequestParams` is plural, but the `executeRequest` method has just one param

nmanu1: I made it plural because there might be multiple parameters. But I understand what you mean that there is only one `data` param, so I'll change it to `RequestDataType`

sample-app/src/components/DropdownSection.tsx
```diff
@@ -12,7 +12,7 @@ export interface DropdownSectionProps {
   optionIdPrefix: string,
   onFocusChange?: (value: string, focusedOptionId: string) => void,
   onLeaveSectionFocus?: (pastSectionEnd: boolean) => void,
-  onClickOption?: (option: Option, optionIndex: number) => void,
+  onSelect?: (optionValue: string, optionIndex: number, ) => void,
```
cea2aj, line 15: nit: Looks like there's an extra comma

sample-app/src/components/DropdownSection.tsx
```diff
@@ -29,12 +29,13 @@ export default function DropdownSection({
...
   onFocusChange = () => {},
   onLeaveSectionFocus = () => {},
-  onClickOption = () => {},
+  onSelect = () => {},
   label = '',
   cssClasses
 }: DropdownSectionProps): JSX.Element | null {
 
   const [focusedOptionIndex, setFocusedOptionIndex] = useState<number>(0);
+  const wasActive = useRef<boolean>(focusStatus === 'active');
```
oshi97, line 38: this is a little out of scope but is focusStatus ever anything but 'active' or 'inactive'? could it just be an `isFocused` boolean?

sample-app/src/components/DropdownSection.tsx
```diff
@@ -59,30 +60,31 @@ export default function DropdownSection({
...
+      decrementOptionFocus();
+    } else if (evt.key === 'Enter') {
+      onSelect(options[focusedOptionIndex].value, focusedOptionIndex);
     }
-  }, [focusStatus, focusedOptionIndex, onFocusChange, optionIdPrefix, options]);
+  }
 
   useEffect(() => {
     if (focusStatus === 'active') {
+      if (!wasActive.current) {
```
oshi97, line 78: why did we need to combine these two effects? I also feel like we could have pass dependencies to the addEventListener effect so we're not adding and removing event listeners unnecessarily

sample-app/src/components/DropdownSection.tsx
nmanu1: good catch. there used to be a third status, but now that there are only two, we can make it a boolean

sample-app/src/components/DropdownSection.tsx
```diff
@@ -59,30 +60,31 @@ export default function DropdownSection({
...
+        onFocusChange(options[focusedOptionIndex].value, `${optionIdPrefix}-${focusedOptionIndex}`);
```
oshi97, line 79: maybe a dumb question but why do we need to call `onFocusChange` in here and also in `increment/decrementOptionFocus`?

sample-app/src/components/DropdownSection.tsx
nmanu1: updated!

sample-app/src/components/DropdownSection.tsx
nmanu1: `increment/decrementOptionFocus` take care of changing what option is focused within the section. when the focus changes (and stays within the section), they call `onFocusChange`. but, when a section that used to be unfocused is focused, that is also a change in focus, which is why we need the `onFocusChange` call in the useEffect

sample-app/src/components/DropdownSection.tsx
nmanu1: By adding the useRef, there wasn't anything differentiating the two useEffects because we're relying on the ref rather than a dependency to determine when the focus has changed. So it was simpler to combine them.

For the event listeners, if we pass dependencies, the other 3 functions (`handleKeyDown`, `increment/decrementOptionFocus`) would need to be wrapped in useCallbacks. I can do that if we prefer to be on the safe side, but right now, this useEffect isn't causing additional adding/removing of the listeners because any time this component would have to re-render, InputDropdown also needs to re-render. And when InputDropdown re-renders, these sections are unmounted and all the children sections are re-mounted anyways.

oshi97 (approved):

sample-app/src/components/FilterSearch.tsx
```diff
@@ -0,0 +1,124 @@
...
+import { useSynchronizedRequest } from "../hooks/useSynchronizedRequest";
+
+const SCREENREADER_INSTRUCTIONS = 'When autocomplete results are available, use up and down arrows to review and enter to select.'
+
+export interface FilterSearchProps {
+  title: string,
+  sectioned: boolean,
+  searchFields: Omit<SearchParameterField, 'fetchEntities'>[],
+  screenReaderInstructionsId: string,
+  cssClasses: {
```
tmeyer2115, line 16: Let's make all these classes optional, in accordance with what Connor's been doing.

sample-app/src/components/DropdownSection.tsx
```diff
@@ -24,17 +24,18 @@ export interface DropdownSectionProps {
...
-  onClickOption = () => {},
+  onSelect = () => {},
   label = '',
   cssClasses
 }: DropdownSectionProps): JSX.Element | null {
 
   const [focusedOptionIndex, setFocusedOptionIndex] = useState<number>(0);
+  const wasFocused = useRef<boolean>(isFocused);
 
   function incrementOptionFocus() {
```
tmeyer2115, line 40: I was looking at the logic for `decrementOptionFocus`. In the `else` condition, we know that `newIndex <= -1`. That results in the following call `onLeaveSectionFocus(false)`. This call reads confusingly. It suggests that we don't leave the section's focus. That's outside the scope of this item, but I think we should fix before we release the sample-app.

sample-app/src/components/DropdownSection.tsx
```diff
@@ -1,18 +1,18 @@
...
 
 export interface DropdownSectionProps {
-  focusStatus?: string,
+  isFocused?: boolean,
   options: Option[],
   optionIdPrefix: string,
   onFocusChange?: (value: string, focusedOptionId: string) => void,
   onLeaveSectionFocus?: (pastSectionEnd: boolean) => void,
-  onClickOption?: (option: Option, optionIndex: number) => void,
+  onSelect?: (optionValue: string, optionIndex: number) => void,
```
tmeyer2115, line 15: Minor nit: could we call this `onSelectOption`? I want to differentiate between selecting a section and an option.

sample-app/src/hooks/useSynchronizedRequest.tsx
```diff
@@ -0,0 +1,24 @@
+import { useRef, useState } from "react";
+
+export function useSynchronizedRequest<RequestDataType, ResponseType>(
```
tmeyer2115, line 3: Can you provide some JSDoc here? I'm not sure what this function does.

sample-app/src/components/FilterSearch.tsx
nmanu1: I thought it was the customCssClasses that were optional? The cssClasses are currently required for `InputDropdown` and other components, so they would have to be required somewhere. Did you mean that I should move the classes back into FilterSearch to use as default css classes and then make this an optional `customCssClasses` prop?

sample-app/src/components/DropdownSection.tsx
nmanu1: Would it be more clear to call it `onChangeSectionFocus`? Or does `onChangeSectionFocus(false)` still read as though the focus shouldn't be changed?

sample-app/src/hooks/useSynchronizedRequest.tsx
nmanu1: Added!

sample-app/src/components/DropdownSection.tsx
tmeyer2115: We can figure it out! Like I said, outside the scope of this item.

sample-app/src/components/FilterSearch.tsx
cea2aj: The `cssClasses` prop (which I'm calling `customCssClasses` ) and the classes themselves should be optional. In my search bar styling PR, I made the css classes of the `InputDropdown` optional, and we should update any other components so that they are optional as well

nmanu1: Updated!

sample-app/src/components/FilterSearch.tsx
```diff
@@ -0,0 +1,124 @@
...
+import { useSynchronizedRequest } from "../hooks/useSynchronizedRequest";
+
+const SCREENREADER_INSTRUCTIONS = 'When autocomplete results are available, use up and down arrows to review and enter to select.'
+
+export interface FilterSearchProps {
+  title: string,
+  sectioned: boolean,
+  searchFields: Omit<SearchParameterField, 'fetchEntities'>[],
+  screenReaderInstructionsId: string,
+  customCssClasses?: {
```
tmeyer2115, line 16: We should follow the example in this file: https://github.com/yext/answers-headless-react/pull/65/files#diff-19c78cfd82fdc3cf822987838108d3efc87b37b12796ddffaecda9a58544dc00. Basically, you'll define an interface for the css classes of this Component, say, `FilterSearchCssClasses`. Then, here, you'll have `customCssClasses?: FilterSearchCssClasses`. 

We'll want to do a similar thing for `InputDropdown` and the Section css classes too.

nmanu1: Updated to have an interface for each component and moved the css classes currently used for styling from `VerticalSearchPage` into `FilterSearch` as the built-in css classes

sample-app/src/components/FilterSearch.tsx
```diff
@@ -37,6 +41,7 @@ export default function FilterSearch ({
   const searchParamFields = searchFields.map((searchField) => {
     return { ...searchField, fetchEntities: false }
   });
+  const cssClasses = builtInCssClasses;
```
tmeyer2115, line 44: We'll need to merge the custom and builtInClasses here.

tmeyer2115 (approved):

sample-app/src/components/FilterSearch.tsx
cea2aj: yep, I hadn't gotten around to merging the `useComposedCssClasses` hook yet (although it's ready for review again)

PR #65, opened 2021-11-30
Search Bar Styling

cea2aj:
Add Tailwind styling to the Search Bar

These built-in styles depend on tailwind being added to the page. I also reworked `renderWithHighlighting` into `renderHighlightedValue` because I think it is clearer.

J=SLAP-1639
TEST=visual

View the styling on the page and visually compare to the mocks


cea2aj:
Screen Shot 2021-11-30 at 2 53 35 PM

sample-app/src/components/DropdownSection.tsx
```diff
@@ -87,7 +87,7 @@ export default function DropdownSection({
 
   function renderOption(option: Option, index: number) {
     const className = classNames(cssClasses.option, {
-      [cssClasses.focusedOption]: focusStatus === 'active' && index === focusedOptionIndex
+      [cssClasses.focusedOption ?? '']: focusStatus === 'active' && index === focusedOptionIndex
```
tmeyer2115, line 90: Could we change this logic? It's a minor nit, but having 
```
{ '': true/false }
```
feels kinda odd. Maybe we don't have this boolean check unless `focusedOption` is defined.

sample-app/src/components/InputDropdown.tsx
```diff
@@ -9,13 +9,18 @@ interface Props {
   screenReaderInstructionsId: string,
   screenReaderText: string,
   onSubmit?: (value: string) => void,
+  renderSearchButton?: () => JSX.Element | null,
+  renderLogo?: () => JSX.Element | null,
+  renderResultIcon?: () => JSX.Element | null,
```
tmeyer2115, line 14: What would you think of moving this prop down to the section-level? Each section could define its own result icons.

sample-app/src/components/SearchBar.tsx
```diff
@@ -1,24 +1,59 @@
...
+  inputDropdownContainer: 'bg-white shadow border rounded-3xl border-gray-300 w-full overflow-hidden',
+  inputElement: 'outline-none flex-grow border-none h-full px-2',
+  logoContainer: 'w-8 mx-2',
+  option: 'flex items-center py-1 px-2 cursor-pointer',
+  resultIconContainer: 'opacity-20 w-8 h-8 pl-1 mr-4',
+  searchButtonContainer: 'w-8 h-full mx-2',
+  submitButton: 'h-full w-full'
+}
+
+interface SearchBarCssClasses {
```
oshi97, line 32: Instead of defining SearchBarCssClasses here, I wonder if we could do something like 
```ts
interface SearchBarCssClasses extends DropdownSectionCssClasses, InputDropdownCssClasses {
//extra classes
}
``` 
Unfortunately I don't think this addresses conflicting names. I had a thought of breaking up SearchBarCssClasses into smaller sub objects i.e
```ts
interface SearchBarCssClasses {
  DropdownSection: DropdownSectionCssClasses,
  InputDropdownCssClasses: InputDropdownCssClasses
}
```
But I don't really like how this bleeds implementation details. Instead of saying specifically
`DropdownSection: DropdownSectionCssClasses`. We could be more general about how we name the sub objects to not bleed details but even then I don't really like the extra complexity. I don't really have a solution just throwing up ideas.

sample-app/src/components/InputDropdown.tsx
cea2aj: sure, that makes sense

sample-app/src/components/SearchBar.tsx
```diff
@@ -76,50 +114,70 @@ export default function SearchBar({
...
+              options={options}
+              optionIdPrefix='Autocomplete__option-0'
+              onFocusChange={value => {
+                answersActions.setQuery(value);
+              }}
+              onClickOption={option => {
+                answersActions.setQuery(option.value);
+                executeQuery();
+              }}
+              cssClasses={cssClasses}
```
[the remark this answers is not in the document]
cea2aj: Yes, that's a downside of this approach that I discussed with Tom when we decided to go this direction. In some situations, we can pass the sub components to the parent components so that the css classes of the sub component is no longer part of the class interface of the parent. The `InputDropdown` is controlled by the `SearchBar` so that approach wouldn't work so great here, but it might work well in other situations. 

I think the manual bookkeeping makes the most sense for now. If there are conflicts, we can namespace the conflicting names for the subcomponents. These class interfaces shouldn't be too large for most components, so I think it will be pretty manageable. If we run into problems with that approach, there are more nuclear solutions such as name-spacing every classname with the module name (e.g. container becomes `SearchBar_container`) or separating the classes object by subcomponent as you suggested

sample-app/src/components/SearchBar.tsx
```diff
@@ -1,24 +1,59 @@
...
+  inputElement: 'outline-none flex-grow border-none h-full px-2',
+  logoContainer: 'w-8 mx-2',
+  option: 'flex items-center py-1 px-2 cursor-pointer',
+  resultIconContainer: 'opacity-20 w-8 h-8 pl-1 mr-4',
+  searchButtonContainer: 'w-8 h-full mx-2',
+  submitButton: 'h-full w-full'
+}
+
+interface SearchBarCssClasses {
+  container?: string,
```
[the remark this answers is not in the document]
cea2aj: I was imaging that we might have something like API extractor setup which will generate docs for the class interfaces. I like the idea of having a default dummy class. I was thinking that could be helpful when looking at the dom

sample-app/src/components/InputDropdown.tsx
cea2aj: This prop actually was actually put in here by mistake. Currently the Options which are passed to the DropdownSection define the rendering of the result icon

sample-app/src/components/SearchBar.tsx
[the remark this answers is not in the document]
cea2aj: Another way to reduce this a bit is to have the SearchBarCssClasses interface extend the interfaces of the subcomponents such as the InputDropdown

sample-app/src/components/SearchBar.tsx
cea2aj: I think the extending makes sense just to clean things up even if it doesn't address the naming conflicting issue

sample-app/src/components/DropdownSection.tsx
cea2aj: I made an update which addresses the nit, however I feel like it's missing the point of the `classNames` library, but let me know if you prefer it

sample-app/src/components/SearchBar.tsx
cea2aj: I updated to extend the DropdownSectionCssClasses and the InputDropdownCssClasses

oshi97 (approved):

PR #66, opened 2021-11-30
SessionId support

yen-tt:
- update AnswersHeadlessProvider to accept a sessionTrackingEnabled field. If true, either 1) get session id from local storage or 2) generate a uuid and store that in local storage. Set the id as sessionId in headless state.

J=SLAP-1726
TEST=manual

- clear all cookies, set sessionTrackingEnabled to true. See that sessionId is tagged in universal and vertical search request. set sessionTrackingEnabled to false and see that sessionId no longer appears
- set sessionTrackingEnabled to true and commented out setSessionId. see that cookie is dropped.
- set sessionTrackingEnabled to true, see that session id is the same in the same tab over reloads. Opening a new tab creates a new session id.


package.json
```diff
@@ -18,7 +18,7 @@
   "dependencies": {
     "@reduxjs/toolkit": "^1.6.2",
     "@types/react": "^17.0.15",
-    "@yext/answers-headless": "^0.1.0-beta.4",
+    "@yext/answers-headless": "file:~/answers-headless",
```
cea2aj, line 21: Just confirming we're gong to publish a new headless version and update this before merging

yen-tt: Yup!

src/AnswersHeadlessProvider.tsx
```diff
@@ -1,19 +1,38 @@
...
+  verticalKey?: string,
+  sessionTrackingEnabled?: boolean
+};
 
 export function AnswersHeadlessProvider(props: Props): JSX.Element {
-  const { children, verticalKey, ...answersConfig } = props;
+  const { children, verticalKey, sessionTrackingEnabled=true, ...answersConfig } = props;
   const answers: AnswersHeadless = provideAnswersHeadless(answersConfig);
   verticalKey && answers.setVerticalKey(verticalKey);
+  handleSessionTracking(answers, sessionTrackingEnabled);
```
cea2aj, line 16: Rather than having `handleSessionTracking`, we could abstract out the window session storage code into a method called `acquireSessionId`. I'm thinking `acquire` rather than `get` because there may be the side effect of setting the session Id, and I think `acquire` captures that better. So the code would look like:

```
answers.setSessionTrackingEnabled(sessionTrackingEnabled);
if (sessionTrackingEnabled) {
    const sessionId = acquireSessionId();
    answers.setSessionId(sessionId);
}

This allows us to avoid having a `handleSessionTracking` function operating at different levels of abstraction. Answers headless is very high-level, and the sessionStorage API is very low level.

Comment (yen-tt) [reply src/AnswersHeadlessProvider.tsx]: sounds good. Yeah `acquire` is better in this context. updated

File: tests/utils/acquireSessionId.test.ts [hunk 3]
```diff
@@ -0,0 +1,37 @@
...
+    window.sessionStorage.clear();
+  });
+
+  it('session id exist in session storage', () => {
+    window.sessionStorage.setItem('sessionId', 'some-uuid-in-storage');
+    expect(acquireSessionId()).toEqual('some-uuid-in-storage');
+  });
+
+  it('session id does not exist in session storage', () => {
+    expect(acquireSessionId()).toEqual('some-uuid-value');
```
cea2aj, line 21: Can we also use `uuidString` here?

cea2aj (approved):

package.json
```diff
@@ -29,6 +29,7 @@
     "@testing-library/react": "^12.1.0",
     "@testing-library/user-event": "^13.2.1",
     "@types/jest": "^27.0.2",
+    "@types/uuid": "^8.3.3",
```
oshi97, line 32: is the uuid package itself missing from the package.json?

PR #67, opened 2021-12-01
Navigation Component Styling

cea2aj:
Add tailwind styling to the navigation component

The mocks didn't cover the menu button and dropdown, so I kept them simple for now. UX may want to make some updates to that

J=SLAP-1639
TEST=visual

Visually compare the component to the mocks


tmeyer2115 (approved):

sample-app/src/components/Navigation.tsx
```diff
@@ -2,18 +2,43 @@ import classNames from 'classnames';
...
+  menuContainer?: string
+}
+
+const builtInCssClasses: NavigationCssClasses = {
+  nav: 'border-b border-gray-200 text-gray-600 flex space-x-6 font-medium pt-4',
+  navLink: 'whitespace-nowrap py-4 px-1 font-medium text-md border-b-2 border-opacity-0 hover:border-gray-300',
+  activeNavLink: 'text-blue-600 border-blue-600 border-b-2 border-opacity-100 hover:border-blue-600',
+  menuButtonContainer: 'relative flex flex-grow justify-end mr-4',
+  menuButton: 'flex flex-row items-center font-medium text-md px-1 border-b-2 border-opacity-0 hover:border-gray-300',
+  menuButton___menuOpen: 'bg-gray-200',
```
oshi97, line 25: is the triple underscore intentional? I would have expected just 2 underscores

cea2aj: In a traditional BEM setup, there would be two dashes, however two dashes isn't valid as an object key without putting it in quotes. We could do that, but then we would also need to change how we access the classNames from `cssClasses.menuButton___menuOpen` to `cssClasses['menuButton--menuOpen']`. I would prefer to avoid having that edge case in the code, but it's an option. A solution that I've seen before is to update BEM from `Block__element--modifier` to `Block_element___modifier`. I was using this convention by putting three underscores before the modifier. However two underscores would also work

oshi97: ok got it! was just wondering if it was a typo, 3 is fine

sample-app/src/components/Navigation.tsx
```diff
@@ -2,18 +2,43 @@ import classNames from 'classnames';
...
 interface LinkData {
   to: string,
   label: string
 }
 
 interface NavigationProps {
-  links: LinkData[]
+  links: LinkData[],
+  customCssClasses?: NavigationCssClasses,
+  compositionMethod?: CompositionMethod
```
yen-tt, line 37: what do you think about `cssCompositionMethod` so it's more specific in terms of what it's use for?

cea2aj: sure, I can address that in another PR

PR #68, opened 2021-12-01
Use case: Guided Search

yen-tt:
Changes:
- Added GuidedSearchPage to display a search bar and prompts component
- Added ProgressBar component
- Added Prompts component that display one prompt at a time (only have StandardPrompt template for now), with navigation button to move to previous/next prompt, or submit and get navigate to a normal universal/vertical search page with results based on the constructed query from the prompts.
- Modify PageRouter to include custom layout for individual pages, and added /prompt route

Notes:
- PageRouter currently takes in a single Layout for all pages. To increase customization flexibility, each page provide its own layout instead.
- Possibly consider adding a callback function in SearchBar so other components may run custom logic for when a search is executed (e.g. navigate to another page on search).

J=SLAP-1734
TEST=manual

### DIFF
--- a/sample-app/src/App.tsx
+++ b/sample-app/src/App.tsx
@@ -5,15 +5,22 @@ import PageRouter from './PageRouter';
 import StandardLayout from './pages/StandardLayout';
 import { AnswersHeadlessProvider } from '@yext/answers-headless-react';
 import { universalResultsConfig } from './universalResultsConfig';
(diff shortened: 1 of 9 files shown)


yen-tt:
https://user-images.githubusercontent.com/36055303/144293440-3f6cf417-d39b-4d32-ae13-94bb54d42d62.mov

PR #70, opened 2021-12-01
useAnswersActions/State error message

yen-tt:
- update useAnswersActions and useAnswersState to check if the value of the answers headless context is undefined, which would indicate that those function calls are invoked outside of AnswersHeadlessProvider.

J=SLAP-1648
TEST=manual&auto

- modify App.tsx to use a component with useAnswersActions and useAnswersState outside of AnswersHeadlessProvider. See error display in window console
- see jest tests passed

### DIFF
--- a/src/useAnswersState.tsx
+++ b/src/useAnswersState.tsx
@@ -10,6 +10,10 @@ export type StateSelector<T> = (s: State) => T;
  */
 export function useAnswersState<T>(stateSelector: StateSelector<T>): T {
   const answers = useContext(AnswersHeadlessContext);
+  if (answers.state === undefined) {
+    throw new Error('Attempt to use AnswersHeadless before it\'s initialized.'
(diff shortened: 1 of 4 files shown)


nmanu1 (approved):

src/useAnswersState.tsx
```diff
@@ -10,6 +10,10 @@ export type StateSelector<T> = (s: State) => T;
  */
 export function useAnswersState<T>(stateSelector: StateSelector<T>): T {
   const answers = useContext(AnswersHeadlessContext);
+  if (answers.state === undefined) {
+    throw new Error('Attempted to use AnswersHeadless before it\'s initialized.'
```
oshi97, line 14: maybe just `Attempted to call useAnswersState() outside an AnswersHeadlessProvider`? I think the first sentence is a little confusing because users might not know that AnswersHeadlessProvider is initializing AnswersHeadless

PR #71, opened 2021-12-01
base Visual Autocomplete

oshi97:
This PR tries to implement Visual Autocomplete in a more flexible
way than a huge array of config options. Externally, VisualAutocompleteSections
can be passed as children into a VisualSearchBar component. This allows the sections
to be organized in many more degrees of freedom than config options could allow, as well
as allowing users to insert whatever other random markup they desire.

Params like universalLimit and restrictedVerticals are calculated from the children passed into
the component, to try to reduce the amount of config options.

Also fixed a (likely unimportant) edge case with near me searches, and refactored some things
to make VisualSearchBar simpler.

J=SLAP-1709
TEST=manual

checked I can tab through the autocomplete options and visual entities

debouncing works as expected

loading state is applied (did not walk through it with product but should be easy to tinker with later on)

test universalLimit is passed correctly

test that searches will perform an autocomplete request to check for near me search intent when an autocomplete option
is selected. previously, only searches that were directly typed in could trigger near me behavior


sample-app/src/components/InputDropdown.tsx
```diff
@@ -14,7 +14,7 @@ interface Props {
   screenReaderInstructions: string,
   screenReaderInstructionsId: string,
   screenReaderText: string,
-  onlyAllowDropdownOptionSubmissions: boolean,
+  onlyAllowDropdownOptionSubmissions?: boolean,
```
tmeyer2115, line 17: Out of scope of your PR, but what does this prop do? The name is somewhat confusing.

sample-app/src/hooks/useNearMeSearch.ts
```diff
@@ -0,0 +1,41 @@
...
+type QueryFunc = () => Promise<void>
+type AutocompleteRef = MutableRefObject<Promise<AutocompleteResponse | undefined> | undefined>
+
+/**
+ * Returns a search action that will handle near me searches, by first checking
+ * for near me intents using an autocomplete request.
+ * You can optionally use the provided ref to store autocomplete responses, to avoid
+ * making unnecessary autocomplete requests.
+ */
+export default function useNearMeSearch(
```
tmeyer2115, line 15: This seems like something @yen-tt wanted to add in a previous PR. We ultimately decided against having a single method with all this responsibility. See the relevant comments here:
- https://github.com/yext/answers-headless-react/pull/57#discussion_r752288311
- https://github.com/yext/answers-headless-react/pull/57#discussion_r752293077

sample-app/src/components/SearchBar.tsx
```diff
@@ -33,16 +30,12 @@ export default function SearchBar({
...
-   * Allow a query search to wait on the response to the autocomplete request right
-   * before the search execution in order to retrieve the search intents
-   */
-  const autocompletePromiseRef = useRef<Promise<AutocompleteResponse | undefined>>();
   const [autocompleteResponse, executeAutocomplete] = useSynchronizedRequest(() => {
     return isVertical
       ? answersActions.executeVerticalAutocomplete()
       : answersActions.executeUniversalAutocomplete();
   });
+  const [ executeQuery, autocompletePromiseRef ] = useNearMeSearch(answersActions, geolocationOptions);
```
tmeyer2115, line 38: It feels strange that a method called `useNearMeSearch` is returning a ref to an auto-complete promise.

sample-app/src/components/VisualAutocomplete/VisualSearchBar.tsx
```diff
@@ -0,0 +1,157 @@
...
+            isLoading={isLoading}
+          />
+        }
+        cssClasses={{
+          dropdownContainer: 'Autocomplete',
+          inputElement: 'SearchBar__input',
+          inputContainer: 'SearchBar__inputContainer'
+        }}
+      >
+        {options.length > 0 && renderDropdownSection()}
```
tmeyer2115, line 118: One thing that feels odd to me is the concept of the `VisualAutocompleteSection`. When I see something like that, I would assume we could use it to represent all pieces of the drop-down experience, including query suggestions. But, query suggestions seems to be its own special flower. Will recent searches have to be a special flower as well, or can we use a Section for that?

sample-app/src/components/VisualAutocomplete/VisualAutocompleteSection.tsx
```diff
@@ -0,0 +1,17 @@
...
+  limit?: number
+}
+
+/**
+ * VisualAutocompleteSection is responsible for providing configuration to VisualAutocompleteEntities,
+ * and allowing users to specify arbitrary VisualAutocomplete layouts.
+ * 
+ * This is done by providing results through props.children, using an FACC (function as a child component).
+ */
+export function VisualAutocompleteSection(_: VisualAutocompleteSectionProps) {
```
tmeyer2115, line 15: I'm not totally sold on the purpose of `VisualAutocompleteSection`. It seems like not much more than a pass-through to `VisualAutocompleteEntities`. Do you anticipate it will do more in the future?

sample-app/src/components/VisualAutocomplete/VisualSearchBar.tsx
tmeyer2115: IMO, if `VisualAutocompleteSection` can only be used for entity previews, we should just invoke the `VisualAutocompleteEntities` Component directly in the `props.children` of the `VisualSearchBar`.

sample-app/src/components/InputDropdown.tsx
oshi97: This is from the FilterSearch pr, it's so that you can only submit options from the autocomplete dropdown, and not any arbitrary input ([I was also confused by it](https://github.com/yext/answers-headless-react/pull/64#discussion_r759509677))

oshi97:
2 rebases later this PR has been moved this to https://github.com/yext/answers-headless-react/pull/77

PR #72, opened 2021-12-02
FilterSearch: one selected filter

yen-tt:
Found an unexpected behavior in FilterSearch, where all previously selected filter search options persist when a new filterSearch option is selected. Update to ensure that only the current selected filter is applied.

J=none
TEST=manual

in people page, perform multiple filter search and select multiple options, see that only the latest option is applied.

### DIFF
--- a/sample-app/src/components/FilterSearch.tsx
+++ b/sample-app/src/components/FilterSearch.tsx
@@ -1,5 +1,5 @@
-import { useState } from "react";
-import { useAnswersActions, FilterSearchResponse, SearchParameterField } from '@yext/answers-headless-react';
+import { useRef, useState } from "react";
+import { useAnswersActions, FilterSearchResponse, SearchParameterField, Filter } from '@yext/answers-headless-react';
 import InputDropdown, { InputDropdownCssClasses } from "./InputDropdown";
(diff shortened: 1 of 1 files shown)


sample-app/src/components/FilterSearch.tsx
```diff
@@ -114,6 +115,10 @@ export default function FilterSearch ({
                 setInput(optionValue);
                 const result = filterSearchResponse?.sections[sectionIndex].results[optionIndex];
                 if (result?.filter) {
+                  if (selectedFilterOptionRef.current) {
+                    answersActions.setFilterOption({ ...selectedFilterOptionRef.current, selected: false });
```
oshi97, line 119: little out of scope but do we need the spread operator for the filter here and in setFilterOption?

yen-tt: yea, setFilterOption expects type `SelectableFilter` (Filter's fields + selected field) so we would need to construct a new object with all the fields from the result's filter and the selected boolean.

oshi97: ah yeah I'm blind >.<

oshi97 (approved):
thanks for fixing!

PR #73, opened 2021-12-02
Multi-filterSearch redirect to result page

yen-tt:
Changes:
- Added two pages: a page with Name, Location, and Language FilterSearch component, with a submit button. When submit, it will redirect to a new result page, containing the same three FilterSearch with the selected options and the corresponding vertical results.
- Added a custom FilterSearch component to accommodate for the button in location filterSearch to get user location and update the input in filtersearch
  - NOTE: getUserLocation provide location coordinates, to get the location name, user would need to user other library, such as react-native-geocoding, with apiKey, to perform the conversion (This is not included in this pr)

Notes:
- Similar to guided search user experience item, may be better if individual page can specify their own layout instead of one layout for entire set of routes in PageRouter.
- Any input changes 
- Headless State currently just group every filters and store as a list. To identify and persist the selected options of each filterSearch component when navigating to another page and displaying the same filter searches, user would need to send state through URL, through react-router history/location.state or some shared context. Not too difficult to implement but they are additional steps.
  - This also require some changes to FilterSearch to get the selected filter option and be able to update the initial input
- Logic that requires input changes (beside initial input) or when to trigger submit to FilterSearch would require copying FilterSearch component and making changes to it. (e.g. a button that gets user location and update filterSearch component input, be able to trigger multiple FilterSearch's setFilterOption call base on a submit button on the page)

J=SLAP-1733
TEST=manual

### DIFF
--- a/sample-app/src/App.tsx
+++ b/sample-app/src/App.tsx
@@ -1,6 +1,8 @@
 import './sass/App.scss';
 import VerticalSearchPage from './pages/VerticalSearchPage';
 import UniversalSearchPage from './pages/UniversalSearchPage';
+import FilterSearchPage from './pages/FilterSearchPage';
+import FilterSearchResultsPage from './pages/FilterSearchResultsPage';
 import PageRouter from './PageRouter';
 import StandardLayout from './pages/StandardLayout';
 import { AnswersHeadlessProvider } from '@yext/answers-headless-react';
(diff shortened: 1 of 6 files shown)


yen-tt:
https://user-images.githubusercontent.com/36055303/144494873-576b7178-fccd-4b61-b5dc-d4347b176c65.mov

yen-tt:
https://user-images.githubusercontent.com/36055303/144508016-bcae8a0c-b831-4ffa-a12e-a31e9c4a2a83.mov

PR #74, opened 2021-12-02
Results Styling

cea2aj:
Add styling to the Universal and Vertical Results Sections and Standard Card

J=SLAP-1639
TEST=visual

Visually inspect the site and compare it to the mocks


sample-app/src/components/UniversalResults.tsx
```diff
@@ -20,24 +30,31 @@ interface AppliedFiltersConfig extends Omit<DecoratedAppliedFiltersConfig, 'appl
 
 interface UniversalResultsProps {
   appliedFiltersConfig?: AppliedFiltersConfig,
-  verticalConfigs: Record<string, VerticalConfig>
+  verticalConfigs: Record<string, VerticalConfig>,
+  customCssClasses?: UniversalResultsCssClasses,
+  compositionMethod?: CompositionMethod
```
yen-tt, line 35: nit: maybe the name for the filed in interface should be `cssCompositionMethod` to be a bit more clear

sample-app/src/components/cards/StandardCard.tsx
```diff
@@ -14,28 +47,62 @@ export interface StandardCardProps extends CardProps {
...
-  const { configuration, result } = props;
+  const { configuration, result, customCssClasses, compositionMethod } = props;
+  const cssClasses = useComposedCssClasses(builtInCssClasses, customCssClasses, compositionMethod);
 
-  return (
-    <div className='StandardCard'>
-      {configuration.showOrdinal && result.index && renderOrdinal(result.index)}
-      <div className='StandardCard__contentWrapper'>
-        {result.name && renderTitle(result.name)}
+  const cta1 = result.rawData.c_primaryCTA as CtaData;
```
yen-tt, line 53: Should we have a typeGuard function to check if it follows CtaData interface (log a message if it doesn't) instead of using force type assertion 'as'?

sample-app/src/components/cards/StandardCard.tsx
```diff
@@ -14,28 +47,62 @@ export interface StandardCardProps extends CardProps {
...
+        <div className={cssClasses.ctaContainer}>
+          {cta1 && renderFirstCTA(cta1)}
+          {cta2 && renderSecondCTA(cta2)}
+        </div>
+      }
+    </>);
+  }
+
+  function renderFirstCTA({ label, link, linkType }: CtaData) {
+    return (
```
yen-tt, line 69: nit: probably don't need parentheses if it's just a single HTML element?

sample-app/src/components/cards/StandardCard.tsx
```diff
@@ -14,28 +47,62 @@ export interface StandardCardProps extends CardProps {
...
+  function renderSecondCTA({ label, link, linkType }: CtaData) {
+    return (
+      <button className={cssClasses.cta2}>{label}</button>
+    )
+  }
+
+  function renderOrdinal(ordinal: number) {
+    return (
+      <div className={cssClasses.ordinal}>
+        {ordinal} -
```
yen-tt, line 83: Seems a bit odd with the dash as part of the renderOrdinal's div. Can we have it outside of it? after this renderOrdinal is called

sample-app/src/sections/StandardSection.tsx
```diff
@@ -1,31 +1,38 @@
...
 
-const StandardSection: SectionComponent = function (props: SectionConfig): JSX.Element | null {
-  const { results, verticalKey, cardConfig, viewMore, header } = props;
-  const latestQuery = useAnswersState(state => state.query.mostRecentSearch); 
+interface StandardSectionCssClasses {
+  section?: string
+}
+
+const builtInCssClasses: StandardSectionCssClasses = {
+  section: ''
```
yen-tt, line 11: just checking, no css stuff for section?

sample-app/src/components/cards/StandardCard.tsx
```diff
@@ -14,28 +47,62 @@ export interface StandardCardProps extends CardProps {
...
-      <div className='StandardCard__ordinal'>
-        {ordinal}
+    <div className={cssClasses.container}>
+      <div className={cssClasses.header}>
+        {configuration.showOrdinal && result.index && renderOrdinal(result.index)}
+        {result.name && renderTitle(result.name)}
       </div>
+      {(result.description ?? cta1 ?? cta2) &&
+        <div className={cssClasses.body}>
+          <div className={cssClasses.descriptionContainer}>
```
yen-tt, line 100: can we add a check for result.description before rendering this div?

sample-app/src/components/UniversalResults.tsx
```diff
@@ -4,14 +4,24 @@ import { DecoratedAppliedFiltersConfig } from '../components/DecoratedAppliedFil
 import SectionHeader from "../sections/SectionHeader";
 import { SectionComponent } from "../models/sectionComponent";
 import { CardConfig } from '../models/cardComponent';
+import { useComposedCssClasses, CompositionMethod } from "../hooks/useComposedCssClasses";
 import classNames from "classnames";
-import '../sass/UniversalResults.scss';
+
+interface UniversalResultsCssClasses {
+  container?: string,
+  results___loading?: string
```
tmeyer2115, line 12: In BEM, I know that the usual syntax is `[block]__[elem]`. Are triple underscores the convention when the block itself (`results`) is modified?

sample-app/src/components/UniversalResults.tsx
```diff
@@ -81,9 +98,14 @@ function renderVerticalSections(props: VerticalSectionsProps): JSX.Element {
         return <SectionComponent
           results={results}
           verticalKey={verticalKey}
-          header={<SectionHeader {...{ label, resultsCountConfig, appliedFiltersConfig }}/>}
+          header={<SectionHeader {...{ 
+            label, 
+            resultsCountConfig,
+            appliedFiltersConfig,
+            verticalKey,
+            viewAllButton: verticalConfig.viewAllButton
```
tmeyer2115, line 106: In the new mocks, is the `viewAllButton` moved to the top of a section?

sample-app/src/components/VerticalResults.tsx
```diff
@@ -17,13 +27,16 @@ interface VerticalResultsDisplayProps {
...
+  const cssClasses = useComposedCssClasses(builtInCssClasses, customCssClasses, compositionMethod);
 
   if (results.length === 0) {
     return null;
   }
 
-  const resultsClassNames = classNames('VerticalResults', { 'VerticalResults--loading': isLoading });
+  const resultsClassNames = cssClasses.results___loading
+    ? classNames({ [cssClasses.results___loading]: isLoading })
+    : '';
```
tmeyer2115, line 39: Are we able to get rid of the `VerticalResults` CSS class because it has no associated styling?

sample-app/src/components/cards/StandardCard.tsx
```diff
@@ -1,11 +1,44 @@
...
+}
+
+export interface StandardCardCssClasses {
+  container?: string,
+  header?: string,
+  body?: string,
+  descriptionContainer?: string,
+  ctaContainer?: string,
+  cta1?: string,
+  cta2?: string,
```
tmeyer2115, line 21: For my own edification, what guidelines do you use for determining what rises to the level of a block vs an element? Would `ctaContainer__cta1` and `ctaContainer__cta2` have worked here?

sample-app/src/components/cards/StandardCard.tsx
cea2aj: Sure

sample-app/src/components/VerticalResults.tsx
cea2aj: Yes, but I could leave it in so that it's customizable

sample-app/src/components/cards/StandardCard.tsx
cea2aj: I'm treating the component as the block, so these represented in BEM would look like: `StandardCard__ctaContainer`, `StandardCard__cta1`, and `StandardCard__cta2`. Therefore everything after the double underscore is an element. And I'm using a triple underscore to indicate modifiers. See the [explanation here](https://github.com/yext/answers-headless-react/pull/67#discussion_r760471334)

sample-app/src/components/cards/StandardCard.tsx
cea2aj: will do

sample-app/src/components/UniversalResults.tsx
cea2aj: I'm using the triple underscore to indicate the modifier since using dashes have some implications for CSS-Modules. See the [explanation here](https://github.com/yext/answers-headless-react/pull/67#discussion_r760471334)

sample-app/src/components/UniversalResults.tsx
cea2aj: yep

sample-app/src/components/VerticalResults.tsx
tmeyer2115: I'm okay getting rid of it! Trim the fat. Drain the swamp.

sample-app/src/components/UniversalResults.tsx
tmeyer2115: Got it! I thought you and Oshi had discussed this somewhere, but I couldn't find the comment.

sample-app/src/components/cards/StandardCard.tsx
cea2aj: these functions are so small, so i'll just move the rendering into the `renderCTAs` function

sample-app/src/components/cards/StandardCard.tsx
cea2aj: We don't have mocks for an ordinal, so is it okay if I just leave it for now?  (Or I could remove the dash). We would will probably update it anyway once we get guidance from UX

yen-tt: gotcha, yeah removing it as fine too

cea2aj: UX says we can simply not render the ordinal for now, and we'll update once we get mocks

sample-app/src/components/cards/StandardCard.tsx
```diff
@@ -14,28 +58,51 @@ export interface StandardCardProps extends CardProps {
...
-        {result.name && renderTitle(result.name)}
-      </div>
-    </div>
-  );
-}
+  const cta1 = isCtaData(result.rawData.c_primaryCTA) ? result.rawData.c_primaryCTA : undefined;
+  const cta2 = isCtaData(result.rawData.c_secondaryCTA) ? result.rawData.c_secondaryCTA : undefined;
+
+  // TODO (cea2aj) We need to handle the various linkType so these CTAs are clickable
+  function renderCTAs(cta1?: CtaData, cta2?: CtaData) {
```
oshi97, line 68: super out of scope but I think it would be cool if we had something like this for cards
https://react-bootstrap.github.io/components/cards/

cea2aj: Ya that would be cool!

sample-app/src/components/cards/StandardCard.tsx
```diff
@@ -41,47 +41,44 @@ interface CtaData {
...
 
+function isCtaData(data: unknown): data is CtaData {
+  if (typeof data !== 'object' || data === null) {
+    return false;
+  }
+  const ctaData = data as CtaData;
+  const expectedKeys = ['label', 'link', 'linkType'];
+  return expectedKeys.every(key => {
+    return key in ctaData;
+  });
```
yen-tt, line 52: can we just do: 
```
const expectedKeys = ['label', 'link', 'linkType'];
  return expectedKeys.every(key => {
    return key in data;
  });
```
to directly check data without the force assertion?

oshi97 (approved):

sample-app/src/components/cards/StandardCard.tsx
```diff
@@ -41,48 +41,47 @@ interface CtaData {
   linkType: string
 }
 
+function isCtaData(data: unknown): data is CtaData {
```
tmeyer2115, line 44: Should this function return a `boolean`? Or does the `is CtaData` syntax tell the IDE/TS the wrapped data is `CtaData` and therefore we get field autocompletion?

sample-app/src/components/cards/StandardCard.tsx
cea2aj: ya, that'll work. I tried that at first but there was an error because typescript couldn't guarantee the data was an object. I've since added a check to make sure it's an object, so the casting is no longer necessary

sample-app/src/components/cards/StandardCard.tsx
cea2aj: Correct, it's a type guard which tells TS whether or not the `data` is a `CtaData`. The underlying javascript has this function simply returning a boolean. You can think of `data is CtaData` as a boolean which TS is able to get some extra info from.

yen-tt (approved):

PR #76, opened 2021-12-03
Automate WCAG testing

yen-tt:
Add automate WCAG testing to sample app

- Move component jest test to be in sample-app instead of top level headless-react repo
  - copied setup folder to sample-app to so the test environment is isolated from the top level testing stuff (and since we may plan on splitting up these two packages?)
- added browser automation (e.g. server, page navigator, page operator) and wcag analyzer test (index.js and test locations). The setup is similar to what is in theme, with some modification:
  -  For single page application, navigate to vertical page through the navigation bar instead from url changes (404 not found). 
  -  Previously, a query is automatically set through URL. Sample app does not support that functionality, so puppeteer must manually click on search bar, type input, and enter.
  - Since class names only include tailwind css names as of now, the selectors used by puppeteer page are html element tags with an assumption that searchbar and navigation component are on top of the page.
- Added github action for WCAG test
  - use Node 15 for npm 7

J=SLAP-1635
TEST=manual & auto

- ran jest tests in sample app, see that everything passed
- ran wcag test in sample app in non-headless mode, see that all the test locations are navigated to and performed proper search. See that wcag results were printed to console.
- see that github action for [wcag test got triggered](https://github.com/yext/answers-headless-react/runs/4413638966?check_suite_focus=true) and the results was logged


sample-app/tests/browser/pagenavigator.js
```diff
@@ -0,0 +1,40 @@
...
+    await this._page.goto(url);
+    await waitTillHTMLRendered(this._page);
+  }
+
+  async click(selector) {
+    await this._page.click(selector);
+    await waitTillHTMLRendered(this._page);
+  }
+
+  async type(input) {
```
tmeyer2115, line 27: The `type` and `search` methods seem outside the responsibility of the `PageNavigator` class. These seem more like actions you can perform on the current page, whereas the other two methods can get you from one page to another. Maybe we should consider putting these methods elsewhere or renaming the class.

sample-app/tests/browser/pageoperator.js
```diff
@@ -0,0 +1,58 @@
...
+ * - name: name of this test
+ * - path: additional path to append to base url
+ * - viewport: viewport of the page (if omit, desktop view is used by default)
+ * - commands: a list of actions to perform on the page by the page navigator
+ */
+class PageOperator {
+  /**
+   * @param {PageNavigator} pageNavigator page navigator
+   * @param {import('puppeteer').Page} page A Pupeteer Page
+   * @param {Object[]} testLocations a list of test locations to navigate to
```
tmeyer2115, line 15: Is there a type for `testLocation`? If not, that could be useful.

sample-app/tests/browser/pageoperator.js
```diff
@@ -0,0 +1,58 @@
...
+    this._page = page;
+    this._testLocations = testLocations;
+    this._testLocationIndex = -1;
+  }
+
+  hasNextTestLocation() {
+    return this._testLocationIndex < this._testLocations.length - 1;
+  }
+
+  async nextTestLocation() {
```
tmeyer2115, line 28: I'm not crazy about the name `TestLocation`. I don't think it accurately describes the object. The name suggests it's essentially just the URL path for the test. I think we should come up with a more descriptive name. As a casual reader of the code, I would imagine this method just moves me to another page. But, it does much more than that.

yen-tt: updated to `testInstruction` instead of `testLocation`

sample-app/tests/browser/pagenavigator.js
yen-tt: Updated:
 PageNavigator => PageOperator (with all these methods still in one class)
 PageOperator => TestOperator (in change of executing testLocation, now call testInstructions)

sample-app/tests/browser/pageoperator.js
yen-tt: Added more descriptive jsdoc for the object properties. No type since these test files are js only right now

sample-app/tests/browser/testoperator.js
```diff
@@ -0,0 +1,57 @@
+const { BrowserPageWidths } = require("./constants");
+
+/**
+ * Responsible for browser navigation based on given test instructions.
+ */
+class TestOperator {
```
tmeyer2115, line 6: The `Operator` suffix makes it sound a bit like this class performs some kind of operations on the tests, rather than just running them. How about `TestRunner`? Does that accurately describe the class's purpose?

sample-app/tests/browser/testoperator.js
```diff
@@ -0,0 +1,57 @@
...
+  /**
+   * @param {PageOperator} pageOperator page operator
+   * @param {import('puppeteer').Page} page A Pupeteer Page
+   * @param {Object[]} testInstructions a list of test instructions to perform
```
tmeyer2115, line 10: Nit: Instead of `TestInstruction` what if we just called it `Test`?

sample-app/tests/browser/testoperator.js
yen-tt: yea, sounds good

sample-app/tests/wcag/index.js
```diff
@@ -38,9 +38,9 @@ async function wcagTester() {
 
   let results = [];
   try {
-    const operator = new TestOperator(pageOperator, page, testInstructions);
-    while (operator.hasNextTestInstruction()) {
-      await operator.nextTestInstruction();
+    const operator = new TestRunner(pageOperator, page, tests);
```
tmeyer2115, line 41: Nit: can we rename the variable here?

tmeyer2115 (approved):

PR #77, opened 2021-12-04
base visual autocomplete

oshi97:
This PR implements the base VisualSearchBar component, aka VisualAutocomplete.

In order for InputDropdown to display content even when no DropdownSection components were
present, I removed the `numSections > 0` check in InputDropdown.

restrictedVerticals will be used in a follow up PR.

J=SLAP-1709
TEST=manual

checked I can tab through the autocomplete options and visual entities

debouncing works as expected

loading state applies opacity to VisualEntities

universalLimit is passed correctly and is reflected in the ui

searches will perform an autocomplete request to check for near me search intent when an autocomplete option
is selected. previously, only searches that were directly typed in could trigger near me behavior

check universal requests for entity perviews are only made if the renderEntityPreview prop is set

### DIFF
--- a/sample-app/src/components/InputDropdown.tsx
+++ b/sample-app/src/components/InputDropdown.tsx
@@ -1,6 +1,7 @@
-import React, { useReducer, KeyboardEvent, useRef, useEffect, useState } from "react"
+import React, { useReducer, KeyboardEvent, useRef, useEffect, useState, FocusEvent } from "react"
 import DropdownSection from "./DropdownSection";
 import ScreenReader from "./ScreenReader";
+import recursivelyMapChildren from './utils/recursivelyMapChildren';
 
 export interface InputDropdownCssClasses {
   dropdownContainer?: string,
@@ -17,7 +18,7 @@ interface Props {
   screenReaderInstructions: string,
   screenReaderInstructionsId: string,
   screenReaderText: string,
-  onlyAllowDropdownOptionSubmissions: boolean,
+  onlyAllowDropdownOptionSubmissions?: boolean,
   onSubmit?: (value: string) => void,
   renderSearchButton?: () => JSX.Element | null,
   renderLogo?: () => JSX.Element | null,
@@ -78,18 +79,20 @@ export default function InputDropdown({
   const [childrenKey, setChildrenKey] = useState(0);
   const [screenReaderKey, setScreenReaderKey] = useState(0);
 
-  const inputRef = useRef<HTMLInputElement>(document.createElement('input'));
-  const dropdownRef = useRef<HTMLDivElement>(document.createElement('div'));
+  const inputRef = useRef<HTMLInputElement>(null);
+  const dropdownRef = useRef<HTMLDivElement>(null);
+  const inputDropdownRef = useRef<HTMLDivElement>(null);
 
   if (!shouldDisplayDropdown && screenReaderKey) {
     setScreenReaderKey(0);
   }
 
-  const childrenArray = React.Children.toArray(children);
-  const childrenWithProps = childrenArray.map((child, index) => {
+  let numSections = 0;
+  const childrenWithProps = recursivelyMapChildren(children, (child, index) => {
     if (!(React.isValidElement(child) && child.type === DropdownSection)) {
       return child;
     }
+    numSections++;
 
     const modifiedOnSelectOption = (optionValue: string, optionIndex: number) => {
       child.props.onSelectOption?.(optionValue, optionIndex);
@@ -113,7 +116,6 @@ export default function InputDropdown({
     }
   });
 
-  const numSections = childrenWithProps.length;
 
   /**
    * Handles changing which section should become focused when focus leaves the currently-focused section.
@@ -148,7 +150,7 @@ export default function InputDropdown({
       evt.preventDefault();
     }
 
-    if (evt.key === 'Escape' || evt.key === 'Tab') {
+    if (evt.key === 'Escape') {
       dispatch({ type: 'HideSections' });
     } else if (evt.key === 'ArrowDown' && numSections > 0 && focusedSectionIndex === undefined) {
       dispatch({ type: 'FocusSection', newIndex: 0 });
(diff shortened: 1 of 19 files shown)


sample-app/src/components/VisualAutocomplete/SampleVisualSearchBar.tsx
```diff
@@ -0,0 +1,78 @@
...
+export default function SampleVisualSearchBar() {
+  return (
+    <VisualSearchBar
+      placeholder='Search...'
+      screenReaderInstructionsId='SearchBar__srInstructions'
+      headlessId='visual-autocomplete'
+      debounceTime={150}
+      renderVisualEntities={(isLoading) => (
+        <div className={isLoading ? 'opacity-50' : ''}>
+          <VisualEntities verticalKey='people'>
```
tmeyer2115, line 17: I'm not sure what's a better name here: `VisualEntities` or `EntityPreviews`. I'd say lets get Product's opinion.

sample-app/src/components/VisualAutocomplete/SampleVisualSearchBar.tsx
```diff
@@ -0,0 +1,78 @@
...
+            {results => (
+              <div className='flex flex-col'>
+                {results.map(r => <FaqCard result={r} key={r.name} />)}
+              </div>
+            )}
+          </VisualEntities>
+        </div>
+      )}
+    >
+      {(autocompleteDropdown, visualEntities) =>
```
tmeyer2115, line 34: Have you seen this pattern elsewhere, where you pass the layout of the Component as `props.children`? I can't remember if it's in the Bootstrap example we talked about. If the spec is unclear, we should as Product how important it is for v1 of VA to have a configurable layout like this. If the answer is "people have to fork the Component" to change the layout, it would make our lives easier.

sample-app/src/components/VisualAutocomplete/VisualSearchBar.tsx
```diff
@@ -0,0 +1,133 @@
...
+    return answersActions.executeUniversalAutocomplete();
+  });
+
+  const [visualAutocompleteState, executeVisualEntitiesQuery] = useVisualEntities(headlessId, debounceTime);
+  const { verticalResultsArray, isLoading: visualEntitiesLoading } = visualAutocompleteState;
+  const autocompleteResults = autocompleteResponse?.results || [];
+  const visualEntities = renderVisualEntities(visualEntitiesLoading, verticalResultsArray);
+  const universalLimit = calculateUniversalLimit(visualEntities);
+
+  function renderDropdownSection() {
```
tmeyer2115, line 66: How about `renderQuerySuggestions`? That's a bit more descriptive.

sample-app/src/components/InputDropdown.tsx
```diff
@@ -78,18 +79,20 @@ export default function InputDropdown({
...
+  const inputDropdownRef = useRef<HTMLDivElement>(null);
 
   if (!shouldDisplayDropdown && screenReaderKey) {
     setScreenReaderKey(0);
   }
 
-  const childrenArray = React.Children.toArray(children);
-  const childrenWithProps = childrenArray.map((child, index) => {
+  let numSections = 0;
+  const childrenWithProps = recursivelyMapChildren(children, (child, index) => {
```
tmeyer2115, line 91: For my own edification, why do we need this recursion here? Does this mean that if we were to render 10 entity previews, say, each would be registered as a "child with props"? Does `childrenWithProps` need to represent each individual Component rendered inside the dropdown as opposed to the actual top-level sections?

oshi97: this is needed so that InputDropdown can handle DropdownSection children that are not direct children, i.e. if you wanted to drop it in a div or fragment
something like
```tsx
<InputDropdown>
  <>
    <DropdownSection/>
  </>
</InputDropdown>
```
with the above example InputDropdown would not see any DropdownSections without recursing through children.
I'm unaware whether the intention was for DropdownSection to only be specified as top level children, but that feels a little restrictive. 
For your 2nd/3rd question, every child of InputDropdown is included in `childrenWithProps`, not just children of type DropdownSection. For instance, in the above example `childrenWithProps` only contains the fragment.

sample-app/src/components/VisualAutocomplete/SampleVisualSearchBar.tsx
oshi97: afaik FACC is typically used for exposing some data to the user, not JSX nodes. It is not included in the react-bootstrap, which seems to exclusively use the compound component pattern instead. Both do a similar thing here - which is to provide some set functionality without prescribing much UI. I think compound component here would be cleaner and also more flexible, but we would probably need to rework DropdownSection.

sample-app/src/components/InputDropdown.tsx
```diff
@@ -230,7 +232,7 @@ export default function InputDropdown({
           : ''
         }
       />
-      {shouldDisplayDropdown &&
+      {shouldDisplayDropdown && !hideDropdown &&
```
tmeyer2115, line 235: Let's mash this

sample-app/src/components/VisualAutocomplete/SampleVisualSearchBar.tsx
```diff
@@ -11,33 +11,26 @@ export default function SampleVisualSearchBar() {
       placeholder='Search...'
       screenReaderInstructionsId='SearchBar__srInstructions'
       headlessId='visual-autocomplete'
-      debounceTime={150}
-      renderVisualEntities={(isLoading) => (
+      entityPreviewsDebouncingTime={150}
+      renderEntityPreviews={isLoading => (
         <div className={isLoading ? 'opacity-50' : ''}>
-          <VisualEntities verticalKey='people'>
+          <EntityPreview verticalKey='people'>
```
tmeyer2115, line 17: EntityPreviews

tmeyer2115 (approved):

PR #79, opened 2021-12-06
Filters Styling

cea2aj:
Add tailwind styling to the StaticFilters and Facets components
 
Add a divider component which is used by both Facets and StaticFilters. I think long term we should refactor the Facets and static Filters so that they are using the same logic for rendering. This would reduce code and styling duplication. I did a small refactor of the `StaticFilters` component so that it now takes in an array of facets. This allows the StaticFilters component to render the divider between each set of Filters.

Add @tailwindcss/forms to style the filter options.

Also comment out the non-styled components in the App.

J=SLAP-1639
TEST=visual

Visually compare the sample app to the mocks


sample-app/src/components/Divider.tsx
```diff
@@ -0,0 +1,19 @@
...
+const builtInCssClasses = {
+  divider: 'w-full h-px bg-gray-200 my-4'
+}
+
+interface Props {
+  customCssClasses?: DividerCssClasses,
+  cssCompositionMethod?: CompositionMethod
+}
+
+export default function Divider({ customCssClasses, cssCompositionMethod }: Props) {
```
tmeyer2115, line 16: I'm wondering if we need a full-blown Component for this. There's not a whole lot to this Component. The main thing it seems to get us is the ability to share Divider styling across Components. Could we instead define and export the default `divider` styling in either `Facets` or `StaticFilters` and import it from the other Component?

cea2aj: Sure, I can move it into StaticFilters

sample-app/package.json
```diff
@@ -8,6 +8,7 @@
     "@pmmmwh/react-refresh-webpack-plugin": "0.4.3",
     "@reduxjs/toolkit": "^1.5.1",
     "@svgr/webpack": "5.5.0",
+    "@tailwindcss/forms": "^0.3.4",
```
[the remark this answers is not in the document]
cea2aj: Ya, I'll switch it

sample-app/src/components/Facet.tsx
```diff
@@ -15,11 +15,41 @@ interface FacetProps extends FacetTextConfig {
...
   defaultExpanded?: boolean,
-  onToggle: onFacetChangeFn
+  onToggle: onFacetChangeFn,
+  customCssclasses?: FacetCssClasses,
+  cssCompositionMethod?: CompositionMethod
+}
+
+interface FacetCssClasses {
+  facetLabel?: string,
+  body?: string,
```
yen-tt, line 25: what is `body` css class for?

sample-app/src/components/StaticFilters.tsx
```diff
@@ -65,21 +71,64 @@ export default function StaticFilters(props: StaticFiltersProps): JSX.Element {
...
-              optionHandler={handleFilterOptionChange}
-            />
+    <div className={cssClasses.container}>
+      {filterConfig.map((filterSet, index) => {
+        const isLastFilterSet = index === filterConfig.length - 1;
+        return <fieldset>
+          <legend className={cssClasses.title}>{filterSet.title}</legend>
+          <div className={cssClasses.optionsContainer}>
+            {filterSet.options.map((option, index) => (
+              <div key={index}>
```
yen-tt, line 81: I think we can remove this div? and move `key={index}` into CheckboxFilter tag `<CheckboxFilter key={index} .../>`

sample-app/src/components/Facet.tsx
cea2aj: oops, I forgot to remove that after refactoring

sample-app/src/components/StaticFilters.tsx
cea2aj: yep, updated!

yen-tt (approved):

PR #80, opened 2021-12-06
Results Count Styling

cea2aj:
Update the component and add styling

J=SLAP-1639
TEST=visual 

Visually compare to the mocks


sample-app/src/components/ResultsCount.tsx
```diff
@@ -1,25 +1,76 @@
...
+    resultsCount,
+    'Results'
+  ];
+   
+  function textSpan(value: string | number, className: string) {
+    return <span className={className}>{value}</span>
+  }
+  
+  return <div className={cssClasses.container}>
+    {messageArray.map((value, index) => {
```
tmeyer2115, line 63: Since we're using a `map` here, does the resultant DOM contain any `ul` or `li` elements?

sample-app/src/components/ResultsCount.tsx
```diff
@@ -1,25 +1,76 @@
...
+      const isLastString = index === messageArray.length - 1;
+      const isNumber = typeof value === 'number';
+      
+      const classes = cssClasses.number
+        ? classNames(cssClasses.text, { [cssClasses.number]: isNumber })
+        : cssClasses.text ?? '';
+
+      return isLastString 
+        ? textSpan(value, classes)
+        : textSpan(value + ' ', classes);
```
tmeyer2115, line 73: Minor nit: I think you could accomplish the same effect by storing the mapped `messageArray` and then returning `mappedArray.join(' ')`. That would save you this conditional here and allow you to remove the `isLastString`. Not a huge deal though.

sample-app/src/components/ResultsCount.tsx
cea2aj: nope, just spans

sample-app/src/components/ResultsCount.tsx
cea2aj: `join` returns a string which won't work here since we need spans

sample-app/src/components/ResultsCount.tsx
```diff
@@ -1,25 +1,72 @@
...
+
+  const spanArray = messageArray.map((value, index) => {
+    const isLastString = index === messageArray.length - 1;
+    const isNumber = typeof value === 'number';
+    
+    const classes = cssClasses.number
+      ? classNames(cssClasses.text, { [cssClasses.number]: isNumber })
+      : cssClasses.text ?? '';
+
+    const spanValue = isLastString ? value : `${value} `;
```
yen-tt, line 66: can we just add the spaces to the strings in messageArray instead? that would simplify this a bit. There may be no space in other languages if we do translation stuff in the future

cea2aj: ya, that works too

sample-app/src/components/ResultsCount.tsx
```diff
@@ -1,25 +1,69 @@
...
+export default function ResultsCount({ customCssClasses, cssCompositionMethod }: Props) {
   const resultsCount = useAnswersState(state => state.vertical?.resultsCount) || 0;
   const resultsLength = useAnswersState(state => state.vertical?.results?.length) || 0;
   const offset = useAnswersState(state => state.vertical?.offset) || 0;
-  return <ResultsCountDisplay resultsCount={resultsCount} resultsLength={resultsLength} offset={offset}/>;
+  const cssClasses = useComposedCssClasses(builtInCssClasses, customCssClasses, cssCompositionMethod);
+  return <ResultsCountDisplay resultsCount={resultsCount} resultsLength={resultsLength} offset={offset} cssClasses={cssClasses}/>;
 }
+
+export function ResultsCountDisplay({
```
yen-tt, line 38: there may be other classes that use ResultsCountDisplay directly (like SectionHeader), what do you think about letting ResultsCount pass in the customCssClass and  ResultsCountDisplay can be the one to call `useComposedCssClasses` to include builtInCssClasses

cea2aj: Sure, I updated this to compose in the display component.

yen-tt (approved):

PR #81, opened 2021-12-06
add restrictVerticals to VisualSearchBar

oshi97:
removed hideDropdown from InputDropdown's dispatch, because doing so
would always use the hideDropdown prop from the previous render instead
of the current one.

J=SLAP-1740
TEST=manual

saw that verticals were restricted
saw that typing in "rose" would always display entity previews
before this would not happen since hideDropdown would always be one render off


sample-app/src/components/VisualAutocomplete/VisualSearchBar.tsx
```diff
@@ -106,7 +107,6 @@ export default function VisualSearchBar({
           onSubmit={executeQuery}
           onInputChange={value => {
             answersActions.setQuery(value);
-            updateEntityPreviews(value);
```
oshi97, line 109: while debugging why the "rose" query wouldn't show any entity previews, I noticed updateEntityPreviews would fire twice for every key stroke. This is because both `onInputChange` and `onInputFocus` are fired whenever somebody types into InputDropdown. entity previews are debounced so you wouldn't notice this just by looking at the networks tab

sample-app/src/components/InputDropdown.tsx
```diff
@@ -70,11 +70,12 @@ export default function InputDropdown({
 }: React.PropsWithChildren<Props>): JSX.Element | null {
   const [{
     focusedSectionIndex,
-    shouldDisplayDropdown,
+    shouldDisplayDropdown: shouldDisplayDropdownWithoutForcedHide,
   }, dispatch] = useReducer(reducer, {
     focusedSectionIndex: undefined,
     shouldDisplayDropdown: false,
   });
+  const shouldDisplayDropdown = shouldDisplayDropdownWithoutForcedHide && !hideDropdown;
```
cea2aj, line 78: I would argue that the true `shouldDisplayDropdown` is the one defined here. Perhaps we could rename the `shouldDisplayDropdown` in the reducer to something else such as `dropdownHidden`. That would let us avoid the rename to `shouldDisplayDropdownWithoutForcedHide`. It might read a little strange to have  `const shouldDisplayDropdown = !dropdownHidden && !hideDropdown;`, however we could add comment explaining that we should only show the dropdown if both the state and props allow it

oshi97: renamed shouldDisplayDropdown in the reducer to dropdownHidden and the hideDropdown prop to forceHideDropdown. it's still a code smell but we do have to plan the next (and hopefully last) refactor of InputDropdown after v1 of the sample app

cea2aj (approved):
🔥

PR #82, opened 2021-12-06
Recent Searches Support

yen-tt:
- Update VisualSearchBar to include support for recent searches using an external library. User can configure the number of searches to track, and whether they want to hide recent searches or not.
- this pr does not include css styling

J=SLAP-1710
TEST=manual

execute multiple queries, and see that recent searches appear in visual search bar.


sample-app/src/components/VisualAutocomplete/VisualSearchBar.tsx
```diff
@@ -50,10 +56,23 @@ export default function VisualSearchBar({
...
+  const recentSearches = useMemo(() => {
+    return hideRecentSearches 
+      ? undefined
+      : new RecentSearches({ limit: recentSearchesLimit })
+  }, [recentSearchesLimit, hideRecentSearches]);
+
+  function executeQuery() {
+    if (!hideRecentSearches) {
+      let input = answersActions.state.query.input;
+      input && recentSearches?.setRecentSearch(input);
```
tmeyer2115, line 72: Do we need the `?` after `recentSearches` here?

sample-app/src/components/VisualAutocomplete/VisualSearchBar.tsx
```diff
@@ -94,6 +113,35 @@ export default function VisualSearchBar({
     />
   }
 
+  function renderRecentSearches() {
+    const options = recentSearches?.getRecentSearches().map(result => {
```
tmeyer2115, line 117: Same question about the necessity of the `?`

sample-app/src/components/VisualAutocomplete/VisualSearchBar.tsx
```diff
@@ -94,6 +113,35 @@ export default function VisualSearchBar({
...
+        )
+      }
+    }) ?? [];
+
+    return <DropdownSection
+      options={options}
+      optionIdPrefix='VisualSearchBar-RecentSearches'
+      onFocusChange={value => {
+        answersActions.setQuery(value);
+        updateEntityPreviews(value);
```
tmeyer2115, line 134: For my own edification, why do we update the entity previews when navigating between recent searches?

sample-app/src/components/VisualAutocomplete/VisualSearchBar.tsx
yen-tt: typescript can't deduce that hideRecentSearches guard the instantiation of recentSearches and yell that recentSearches `is possibly 'undefined'`

sample-app/src/components/VisualAutocomplete/VisualSearchBar.tsx
yen-tt: similar as above comment, recentSearches may be undefined if hideRecentSearches is true

sample-app/src/components/VisualAutocomplete/VisualSearchBar.tsx
yen-tt: ah no, we don't actually need the onFocusChange function here. It's removed

sample-app/src/components/VisualAutocomplete/VisualSearchBar.tsx
```diff
@@ -50,10 +57,27 @@ export default function VisualSearchBar({
...
+      return undefined;
+    } else {
+      return new RecentSearches({ limit: recentSearchesLimit, namespace: DEFAULT_RECENT_SEARCH_KEY });
+    }
+  }, [recentSearchesLimit, hideRecentSearches]);
+  const haveRecentSearches = !hideRecentSearches && recentSearches?.getRecentSearches() && recentSearches.getRecentSearches().length > 0;
+
+  function executeQuery() {
+    if (!hideRecentSearches) {
+      let input = answersActions.state.query.input;
```
oshi97, line 76: can be const instead of let

sample-app/src/components/VisualAutocomplete/VisualSearchBar.tsx
```diff
@@ -50,10 +57,27 @@ export default function VisualSearchBar({
   const answersActions = useAnswersActions();
   const query = useAnswersState(state => state.query.input) ?? '';
   const isLoading = useAnswersState(state => state.searchStatus.isLoading) ?? false;
-  const [executeQuery, autocompletePromiseRef] = useSearchWithNearMeHandling(answersActions)
+  const [executeQueryWithNearMeHandling, autocompletePromiseRef] = useSearchWithNearMeHandling(answersActions)
   const [autocompleteResponse, executeAutocomplete] = useSynchronizedRequest(async () => {
     return answersActions.executeUniversalAutocomplete();
   });
+  const recentSearches = useMemo(() => {
```
oshi97, line 64: is it possible to extract recentSearches to a hook?

sample-app/src/components/VisualAutocomplete/VisualSearchBar.tsx
```diff
@@ -50,10 +57,27 @@ export default function VisualSearchBar({
...
+    if (hideRecentSearches) {
+      localStorage.removeItem(DEFAULT_RECENT_SEARCH_KEY);
```
oshi97, line 66: I think we probably don't want to put side effects inside useMemo. Is it necessary to remove recent searches data from local storage?

oshi97: wondering if maybe it makes sense to use session storage, maybe rose would know better how long we want recent searches to persist?

sample-app/src/components/VisualAutocomplete/VisualSearchBar.tsx
```diff
@@ -40,7 +45,9 @@ export default function VisualSearchBar({
   placeholder,
   screenReaderInstructionsId,
   headlessId,
+  hideRecentSearches=false,
```
oshi97, line 48: I think it might be unnecessary to default to false?

sample-app/src/components/VisualAutocomplete/VisualSearchBar.tsx
```diff
@@ -40,7 +45,9 @@ export default function VisualSearchBar({
...
   renderEntityPreviews,
+  recentSearchesLimit=5,
```
oshi97, line 50: nit: think I like spaces around the equals sign

sample-app/src/components/VisualAutocomplete/VisualSearchBar.tsx
```diff
@@ -12,9 +12,12 @@ import useSearchWithNearMeHandling from '../../hooks/useSearchWithNearMeHandling
 import { builtInCssClasses, SearchBarCssClasses } from '../SearchBar';
 import { CompositionMethod, useComposedCssClasses } from '../../hooks/useComposedCssClasses';
 import { ReactComponent as YextLogoIcon } from '../../icons/yext_logo.svg';
+import { ReactComponent as RecentSearchIcon } from '../../icons/history.svg';
 import renderAutocompleteResult from '../utils/renderAutocompleteResult';
+import RecentSearches from "recent-searches";
 
 const SCREENREADER_INSTRUCTIONS = 'When autocomplete results are available, use up and down arrows to review and enter to select.'
+const DEFAULT_RECENT_SEARCH_KEY = '__yxt_recent_searches__'
```
oshi97, line 20: default implies this can be overridden but since it can't, maybe just RECENT_SEARCHES_KEY?

sample-app/src/components/VisualAutocomplete/VisualSearchBar.tsx
```diff
@@ -50,10 +57,27 @@ export default function VisualSearchBar({
...
   });
+  const recentSearches = useMemo(() => {
+    if (hideRecentSearches) {
+      localStorage.removeItem(DEFAULT_RECENT_SEARCH_KEY);
+      return undefined;
+    } else {
+      return new RecentSearches({ limit: recentSearchesLimit, namespace: DEFAULT_RECENT_SEARCH_KEY });
+    }
+  }, [recentSearchesLimit, hideRecentSearches]);
+  const haveRecentSearches = !hideRecentSearches && recentSearches?.getRecentSearches() && recentSearches.getRecentSearches().length > 0;
```
oshi97, line 72: does this work? just removing the last &&
```ts
const haveRecentSearches = !hideRecentSearches && recentSearches?.getRecentSearches()?.length > 0
```

sample-app/src/components/VisualAutocomplete/VisualSearchBar.tsx
```diff
@@ -94,6 +118,31 @@ export default function VisualSearchBar({
     />
   }
 
+  function renderRecentSearches() {
+    const options = recentSearches?.getRecentSearches().map(result => {
+      return {
+        value: result.query,
+        display: (
+          <div className='flex items-center'>
```
oshi97, line 126: would it be possible to use `renderAutocompleteResult` here?

sample-app/src/components/VisualAutocomplete/VisualSearchBar.tsx
yen-tt: In the case the user switch hideRecentSearches from false to true, recent searches may no longer track new queries but the old ones still exists in local storage. Based on the spec regarding privacy concerns, we should clear those data. And when hideRecentSearches is switch back to false, they would have a fresh history. I could pull the remove call out of useMemo

sample-app/src/components/VisualAutocomplete/VisualSearchBar.tsx
yen-tt: nope, attempted that before and it error out for comparing undefined with a number

sample-app/src/components/VisualAutocomplete/VisualSearchBar.tsx
oshi97: if recentSearches is null should we just short circuit and return null? Otherwise we still render a DropdownSection just with an empty options array

sample-app/src/components/VisualAutocomplete/VisualSearchBar.tsx
oshi97: sgtm!

sample-app/src/components/VisualAutocomplete/VisualSearchBar.tsx
oshi97: maybe `length !== 0` instead of `> 0`? alternatively I wouldn't mind breaking this up into two lines

sample-app/src/components/VisualAutocomplete/VisualSearchBar.tsx
yen-tt: renderAutocompleteResult expects result to be of type AutocompleteResult and have a fix magnifying glass icon. I will try to make it more generic and see how it goes

oshi97: I forgot we were doing the whole renderHighlightedValue thing, so if it's annoying it's probably not worth it

sample-app/src/components/VisualAutocomplete/VisualSearchBar.tsx
```diff
@@ -29,6 +32,8 @@ interface Props {
   // The debouncing time, in milliseconds, for making API requests for entity previews
   entityPreviewsDebouncingTime: number,
   renderEntityPreviews?: RenderEntityPreviews,
+  hideRecentSearches?: boolean,
```
cea2aj, line 34: nit: I have a slight preference for naming boolean flags without using negation. In that case, this would be flipped to 'showRecentSearches', and by default the value would be true. I think it's a tad easier to reason about

yen-tt: this is what the spec have, but I can ask rose about it

sample-app/src/components/VisualAutocomplete/VisualSearchBar.tsx
yen-tt: decided not to use it at the end

sample-app/src/components/VisualAutocomplete/VisualSearchBar.tsx
yen-tt: update: pr now use hook

sample-app/src/components/VisualAutocomplete/VisualSearchBar.tsx
yen-tt: sure thing! updated

sample-app/src/icons/history.svg
```diff
@@ -0,0 +1,3 @@
+<svg width="14" height="15" viewBox="0 0 14 15" fill="none" xmlns="http://www.w3.org/2000/svg">
```
cea2aj, line 1: In the component styling items, I removed the width and height from the SVGs I used and added a container around the SVG so that the sizing could be controlled with CSS. We may want to do that here as well to be consistent

yen-tt: i see, updated!

sample-app/src/components/VisualAutocomplete/VisualSearchBar.tsx
```diff
@@ -9,12 +9,28 @@ import { processTranslation } from '../utils/processTranslation';
...
 import { CompositionMethod, useComposedCssClasses } from '../../hooks/useComposedCssClasses';
 import { ReactComponent as YextLogoIcon } from '../../icons/yext_logo.svg';
+import { ReactComponent as RecentSearchIcon } from '../../icons/history.svg';
 import renderAutocompleteResult from '../utils/renderAutocompleteResult';
+import useRecentSearches from '../../hooks/useRecentSearches';
 
 const SCREENREADER_INSTRUCTIONS = 'When autocomplete results are available, use up and down arrows to review and enter to select.'
+const builtInCssClasses: VisualSearchBarCssClasses = { 
+  ...builtInSearchBarCssClasses, 
+  ...{
```
oshi97, line 22: we can get rid of the spread here

sample-app/src/components/VisualAutocomplete/VisualSearchBar.tsx
```diff
@@ -50,10 +70,28 @@ export default function VisualSearchBar({
   const answersActions = useAnswersActions();
   const query = useAnswersState(state => state.query.input) ?? '';
   const isLoading = useAnswersState(state => state.searchStatus.isLoading) ?? false;
-  const [executeQuery, autocompletePromiseRef] = useSearchWithNearMeHandling(answersActions)
+  const [executeQueryWithNearMeHandling, autocompletePromiseRef] = useSearchWithNearMeHandling(answersActions)
   const [autocompleteResponse, executeAutocomplete] = useSynchronizedRequest(async () => {
     return answersActions.executeUniversalAutocomplete();
   });
+  const [
```
oshi97, line 77: it feels a little weird that we return an instance of the recent-searches library and not just the recent searches data object itself? If we imagine that we're not using any particular library for this, it feels a little arbitrary why setRecentSearches() and getRecentSearches() are grouped in an object, but we also provide 2 additional methods after that. I would probably shoot for something as close to a useState hook as possible.
```ts
const [ recentSearches, setRecentSearches, clearRecentSearches ] = useRecentSearches(limit);
```
or maybe even this, which would let us also factor out the effect too
```ts
const [ recentSearches, setRecentSearches ] = useRecentSearches(limit, disableRecentSearches);
```
@tmeyer2115 do you think it would be okay to pass in a disableRecentSearches boolean? It's a little weird, but I also think having a createNewRecentSearchesStorage method is also a little weird?

tmeyer2115: I'm not crazy about the current method signature. I think the two `*Storage` methods are leaking implementation details. They also don't make the method feel particularly "hook" like as Oshi said. I'm also not crazy about adding a boolean to `useRecentSearches`. 

What about making a wrapper class over RecentSearches that has the following methods:
- get(...)
- set(...)
- clear()

The first two would just be pass-throughs. The second would clear out the corresponding bit of local storage. The VA Component could invoke `clear()` whenever `hideRecentSearches` is true. I think having a wrapper also gives us the benefit of not leaking implementation details of `RecentSearch` into other classes. We could make `RecentSearches` a singleton if necessary. I'm not saying this is the best way, but I'd live to avoid adding the boolean to `useRecentSearches` and exposing implementation details about local storage. Happy to discuss further. @oshi97 @yen-tt

tmeyer2115: That interface sounds good to me.

oshi97 (approved):
looks great!

PR #83, opened 2021-12-08
Vertical Links Support

yen-tt:
Added Vertical Links Support in VisualSearchbar

- consider vertical links as an option when constructing the list of options to pass to DropdownSection
  - move onSelectOption to onClick as part of Option interface, so link and normal option can defined its own behavior when click on or pressed enter. 
  - updated renderAutocompleteResults to take in css classes and icon to accommodate for different types of option
  - added css styling

- other fixes:
  - Enter event handler in inputDropdown should only run when the target is the input element.
  -  any react update should be avoided when component get unmounted (i.e. browser navigation or async function related). Added guard in useEntityPreview and InputDropdown

J=SLAP-1712
TEST=manual

used mock data for autocompleteResults[].verticalKeys field:
- See that the vertical links are display under each option appropriately.
- See that click/enter on a link navigate to the right vertical search page with the selected query in search bar
- See that keyboard up/down navigation works properly with and without vertical links in InputDropdown


sample-app/src/components/VisualAutocomplete/VisualSearchBar.tsx
```diff
@@ -111,9 +116,25 @@ export default function VisualSearchBar({
...
+      });
       return {
         value: result.value,
-        display: renderAutocompleteResult(result, cssClasses.resultIconContainer)
+        verticalLinks,
+        render: (onClick: () => void, isOptionFocus: boolean, focusLinkIndex: number) => 
+          renderAutocompleteResult(
+            result,
+            onClick,
+            cssClasses as AutocompleteResultCssClasses,
```
oshi97, line 133: do we need this assertion?

sample-app/src/components/VisualAutocomplete/VisualSearchBar.tsx
```diff
@@ -111,9 +116,25 @@ export default function VisualSearchBar({
       return null;
     }
     const options = autocompleteResults.map(result => {
+      const verticalKeys = hideVerticalLinks ? undefined : ['people', 'events'];//result.verticalKeys;
```
oshi97, line 119: this is from testing right? If we want to keep it we can have a more descriptive comment

sample-app/src/components/VisualAutocomplete/SampleVisualSearchBar.tsx
```diff
@@ -12,6 +13,12 @@ export default function SampleVisualSearchBar() {
       screenReaderInstructionsId='SearchBar__srInstructions'
       headlessId='visual-autocomplete'
       entityPreviewsDebouncingTime={150}
+      verticalKeyToNameMapping={
```
oshi97, line 16: maybe we could have a `handleVerticalLink` prop instead? not sure if it's trying too hard to be generic, but I think it's a little more obvious what the prop is for.

sample-app/src/components/DropdownSection.tsx
```diff
@@ -1,16 +1,22 @@
...
+
+export interface VerticalLink {
+  label: string,
+  verticalKey: string
+}
 
 export interface Option {
   value: string,
-  display: JSX.Element
+  verticalLinks?: VerticalLink[],
```
oshi97, line 11: Ideally, I don't think DropdownSection should know what a VerticalLink is. Is it possible to move `onSelectOption` here so we can have different behavior per option? Then maybe we can get rid of `focusedLinkIndex`. Right now increment/decrementOptionFocus are pretty confusing.

sample-app/src/components/VisualAutocomplete/VisualSearchBar.tsx
yen-tt: nope, forgot to remove when refactoring the styling.

sample-app/src/components/DropdownSection.tsx
yen-tt: if would be a separate function `onFocusOption` that DropdownSection would need, but focusedLinkIndex is needed so DropdownSection can remember which link is in focus instead of the actual option. Alternatively we can have an AutocompleteResult component instead of a render function so it can store its own focus option/link index state and handle its own click/enter event. I will try to see if that would work, otherwise I can refactor increment/decrementOptionFocus into smaller functions

sample-app/src/components/VisualAutocomplete/SampleVisualSearchBar.tsx
yen-tt: so `handleVerticalLink` would transform a given list of vertical keys to a VerticalLink object? it would work

yen-tt: updated

sample-app/src/components/DropdownSection.tsx
yen-tt: seem like we can remove link stuff from dropdownSection unless we want to adopt the same model of inputDropdown + DropdownSection model for AutocompleteOption component to be in charge of navigating the option + links

cea2aj: The InputDropdown is currently responsible for determining which section is focused, and each dropdown section is responsible for maintaining the focus within that section. Having the AutocompleteResult maintain it's own focus is doable-but that's a lot of additional complexity. Instead, what do you think about having a `DropdownSectionWithSubOptions` and keeping `DropdownSection` but renaming it to `SimpleDropdownSection`?

sample-app/src/components/DropdownSection.tsx
```diff
@@ -50,6 +76,28 @@ export default function DropdownSection({
     setFocusedOptionIndex(newIndex);
   }
 
+  function updateToPrevFocusElement() {
+    if(!decrementVerticalLinkFocus()) {
+      decrementOptionFocus();
+    }
+  }
+
+  function decrementVerticalLinkFocus(): boolean {
```
cea2aj, line 85: can we document what this boolean indicates?

sample-app/src/components/DropdownSection.tsx
```diff
@@ -37,8 +43,28 @@ export default function DropdownSection({
...
 
+  function updateToNextFocusElement() {
+    if(!incrementVerticalLinkFocus()) {
+      incrementOptionFocus();
+    }
+  }
+
+  function incrementVerticalLinkFocus(): boolean {
+    const verticalLinks = options[focusedOptionIndex]?.verticalLinks;
+    if (verticalLinks) {
```
cea2aj, line 58: nit: I have a preference for returning early so that the code isn't nested so much. For example, we could have `if (!verticalLinks) { return false; }`, and then the rest of the function would follow without needing to be nested

sample-app/src/components/DropdownSection.tsx
cea2aj: It looks like it's indicating whether or not the decrement was successful. If so, can that be a separate function such as `canDecrementVerticalLinkFocus`?

sample-app/src/components/DropdownSection.tsx
```diff
@@ -67,11 +115,17 @@ export default function DropdownSection({
...
+      updateToNextFocusElement();
     } else if (evt.key === 'ArrowUp' || evt.key === 'ArrowLeft') {
-      decrementOptionFocus();
+      updateToPrevFocusElement();
     } else if (evt.key === 'Enter') {
       onSelectOption(options[focusedOptionIndex].value, focusedOptionIndex);
+      if (focusedLinkIndex !== -1) {
+        const verticalLinks = options[focusedOptionIndex]?.verticalLinks;
+        if (verticalLinks) {
+          browserHistory.push(`/${verticalLinks[focusedLinkIndex].verticalKey}`);
```
cea2aj, line 126: Can you explain why we're modifying browserHistory here?

sample-app/src/components/utils/renderAutocompleteResult.tsx
```diff
@@ -1,19 +1,73 @@
...
-      <MagnifyingGlassIcon />
-    </div>
-    <div>
-      {renderHighlightedValue(result)}
-    </div>
-  </>
+export default function renderAutocompleteResult(
+  result: AutocompleteResult,
+  onClick: () => void,
+  cssClasses: AutocompleteResultCssClasses,
```
cea2aj, line 34: Can we make this optional and set a default of `{}`?

sample-app/src/components/utils/renderAutocompleteResult.tsx
```diff
@@ -1,19 +1,73 @@
...
+  isOptionFocus?: boolean,
+  verticalLinks?: VerticalLink[],
+  focusLinkIndex?: number
+) {
+  const hasFocusLink = focusLinkIndex !== undefined && focusLinkIndex !== -1;
+  const OptionCssClasses = cssClasses.focusedOption
```
cea2aj, line 40: Can we lowercase `OptionCssClasses`?

sample-app/src/components/VisualAutocomplete/VisualSearchBar.tsx
```diff
@@ -134,24 +159,30 @@ export default function VisualSearchBar({
   }
 
   function renderRecentSearches() {
-    const options = recentSearches?.map(result => {
+    const options: Option[] = recentSearches?.map(result => {
       return {
         value: result.query,
-        display: (
-          <div className={cssClasses.recentSearchesOption}>
+        render: (onClick, isOptionFocus) => {
```
cea2aj, line 165: nit: What do you think of `isOptionFocused`?

sample-app/src/components/DropdownSection.tsx
yen-tt: it seems a bit odd to have separate function, since a false may mean decrement/set focused link index to -1 or to the last link index of previous option but still require decrementOptionFocus. But yea I will add documentation to explain the boolean

sample-app/src/components/DropdownSection.tsx
yen-tt: I see, will update!

sample-app/src/components/DropdownSection.tsx
yen-tt: it's from useHistory hook in React Router and we need to modify it like so to navigate to a different page programmatically in the router.

sample-app/src/components/DropdownSection.tsx
oshi97: can't each link be its own option? do we need this idea of sub-options?

yen-tt: That could work, I will try it out

cea2aj: hmm, ya that might be simpler

sample-app/src/components/VisualAutocomplete/SampleVisualSearchBar.tsx
```diff
@@ -12,6 +13,12 @@ export default function SampleVisualSearchBar() {
       screenReaderInstructionsId='SearchBar__srInstructions'
       headlessId='visual-autocomplete'
       entityPreviewsDebouncingTime={150}
+      handleVerticalLinks={verticalKeys => {
```
oshi97, line 16: I was thinking handleVerticalLink would be responsible for changing the page to the correct vertical, that way DropdownSection doesn't have to know about react router

sample-app/src/components/DropdownSection.tsx
yen-tt: I can try out the separate section idea. I don't think the link can be its own option? beside the different styling, the link require the corresponding option's value for input update, which is different from the link's label on display, and the click/enter behavior is different. unless we make a `Option` and `OptionLink` and some interface that merges it maybe

sample-app/src/components/VisualAutocomplete/SampleVisualSearchBar.tsx
yen-tt: we would need some sort of mapping still to get the right label to display in the dropdown beside navigation

oshi97: Ohh I missed that. Maybe this prop can just be for transforming a vertical key to the right label? feel like this function taking in an array of keys is a little arbitrary
```ts
verticalKeyToLabel={{verticalKey => { 
  const label = universalResultsConfig[verticalKey]?.label ?? verticalKey;
  return { label, verticalKey }
}}
```

sample-app/src/components/DropdownSection.tsx
yen-tt: update: yea having it as its own option works!

sample-app/src/components/VisualAutocomplete/SampleVisualSearchBar.tsx
yen-tt: sounds good, updated!

sample-app/src/components/InputDropdown.tsx
```diff
@@ -85,37 +85,67 @@ export default function InputDropdown({
   const inputRef = useRef<HTMLInputElement>(null);
   const dropdownRef = useRef<HTMLDivElement>(null);
   const inputDropdownRef = useRef<HTMLDivElement>(null);
-
   if (!shouldDisplayDropdown && screenReaderKey) {
     setScreenReaderKey(0);
   }
 
+  const isMounted = useRef<boolean>(false);
+  useEffect(() => {
```
oshi97, line 93: Was this needed by something added by this PR or was always an issue with InputDropdown? Also do you know in what kinds of situations we get state updates on unmounted components?

sample-app/src/hooks/useEntityPreviews.tsx
```diff
@@ -25,12 +25,24 @@ export function useEntityPreviews(headlessId: string, debounceTime: number):[ En
       headlessId
     });
   }
+  const isMounted = useRef<boolean>(false);
+  useEffect(() => {
```
oshi97, line 29: same questions here (I assume the two are linked but maybe not...?)

sample-app/src/components/InputDropdown.tsx
yen-tt: I believe a component unmount when the parent component is no longer rendered or the parent component performs an update that does not render this instance (or calling ReactDOM.unmountComponentAtNode). 
This would be a potential issue if a user's custom component that uses InputDropdown does some custom render logic or include some special logic in `option.onClick?.();` the would trigger unmount before calling the following state update functions. 
In the case of this PR, `option.onClick?.();` happens to perform a route navigation necessary for a vertical link option, so the following state update functions would be attempting to update state for an unmounted component.

sample-app/src/hooks/useEntityPreviews.tsx
yen-tt: similar to my response above, `useDebouncedFunction` invoke an async function that would wait for a response, before calling state update functions. There would be instances where during the wait for the response, the component unmounted, then it would proceed to execute the state update functions `setVerticalResultsArray` `setLoadingState` which would cause an error.

sample-app/src/components/FilterSearch.tsx
```diff
@@ -42,7 +42,7 @@ export default function FilterSearch ({
   const searchParamFields = searchFields.map((searchField) => {
     return { ...searchField, fetchEntities: false }
   });
-  const cssClasses = { builtInCssClasses, ...customCssClasses };
+  const cssClasses = { ...builtInCssClasses, ...customCssClasses };
```
cea2aj, line 45: Can we update this to use the `useComposedCssClasses` hook instead?

sample-app/src/components/DropdownSection.tsx
cea2aj: When I prototyped this I used a 'Link' from 'react-router-dom'. Could that work instead? (See: https://github.com/yext-pages/verizon-demo/pull/37/files)

yen-tt: A Link would only work when click on (and then and routes to whatever 'to' field points to), but to navigate programmatically, like when user hit enter on the link option, I believe I tried that before and it didn't work, so I ended up using the router history

sample-app/src/hooks/useEntityPreviews.tsx
oshi97: are universal searches firing when selecting a vertical link? let's move this isMounted handling into useDebouncedSearch, since it seems logical that a debounced function might result in async behavior that we would need this handling for.

sample-app/src/components/InputDropdown.tsx
oshi97: would moving the onClick after setLatestUserInput and dispatch help?

yen-tt: I think it would help? The error appears sporadically, I tested it a couple times with the order swapped and i'm not seeing the error at the moment so it seems fine. However, if I remember correctly, the setState/dispatch stuff does not guarantee unsubscriptions until next dispatch according to some caveat in redux source code so I think there's a chance it could still cause an error? which is why I kept it in this format and a bit hesitant in just swapping the order.

sample-app/src/hooks/useEntityPreviews.tsx
yen-tt: it's more so an edge case/race condition thing. A vertical link is under an option that would display entity preview. So when user navigate with key, this could happen:
1) arrow down => focus on an option (trigger universal search for entity preview)
2)arrow down + enter => select a vertical link and navigate (async done waiting and start executing the other state update functions)
I can't move it into useDebouncedSearch because the issue is within the function that's pass into the useDebouncedSearch. Since the state update occurs in useEntityPreview, I think it makes since to have it in here.  useDebouncedSearch is not aware of those state updates and is only in charge of debounce stuff

sample-app/src/components/InputDropdown.tsx
oshi97: was the stack trace you were getting pointing to anything in redux? I thought that the dispatch and setLatestUserInput here would have nothing (directly) to do with our redux store

yen-tt: the stack trace doesn't have point related to it. It was similar to the one we encountered before:
Screen Shot 2021-12-09 at 3 14 28 PM

sample-app/src/hooks/useEntityPreviews.tsx
oshi97: Did a little research and this warning has actually been removed from the react codebase, so it should probably be gone in react 18?
https://medium.com/doctolib/react-stop-checking-if-your-component-is-mounted-3bb2568a4934
[the linked pr](https://github.com/facebook/react/pull/22114)
It makes me feel a little better that a lot of other people run into this issue, and that our case here is considered a false positive (I think) in terms of it being a memory leak. 
I'm fine with either extracting isMounted into it's own hook, or just letting this errors go unchecked if we're sure they're not memory leaks (which I'm pretty sure they're not...?)

oshi97: It will probably be at least a few months until react 18 so leaning towards just hiding the errors with a hook, and then when react 18 finally comes it should be easy to find usages of the hook

sample-app/src/components/InputDropdown.tsx
yen-tt: update: I swapped the order since I cannot replicate the error anymore and assuming that react state update subscription doesn't have the same caveat as redux.

sample-app/src/components/FilterSearch.tsx
cea2aj: sure

cea2aj (approved):

sample-app/src/components/DropdownSection.tsx
```diff
@@ -3,14 +3,15 @@ import { useState, useEffect, useRef } from "react";
 
 export interface Option {
   value: string,
+  onClick: () => void,
```
oshi97, line 6: this is called on click here, but it also runs when the option is selected with the arrows keys right?

sample-app/src/hooks/useComponentMountStatus.tsx
```diff
@@ -0,0 +1,11 @@
+import { useEffect, useRef } from "react";
+
+export default function useComponentMountStatus() {
```
oshi97, line 3: Can we add a comment saying that this is used to suppress react errors that do not need to be suppressed in the next version of react, and include a link to the react pr?

oshi97 (approved):

PR #84, opened 2021-12-08
Use answers-js-docs answers experience

cea2aj:
Update the sample app to use the answers-js-docs answers experience

Update to use the new experience by creating one page per vertical. I also refactored some common code into a `usePage` hook to reduce code duplication across pages. 

There appears to be an issue were the verticalKey isn't set properly when navigating between pages, however this issue existed before this PR which suggests that it's not related to the `usePage` refactor.

I also updated the key in the AppliedFilters so that the same filter value of different filter types won't cause a React error. 

When FilterSearch and Facets are both on the page, the applied filters component will render both of the filters. I think we should update the app so that just one of them is rendered, but this is out of scope.

J=SLAP-1742
TEST=manual

Load the page and see results from the new experience


sample-app/src/App.tsx
```diff
@@ -1,29 +1,42 @@
...
 
 export default function App() {
   return (
     <AnswersHeadlessProvider
-      apiKey='2d8c550071a64ea23e263118a2b0680b'
-      experienceKey='slanswers'
+      apiKey='3517add824e992916861b76e456724d9'
+      experienceKey='answers-js-docs'
       locale='en'
       verticalKey='people'
```
tmeyer2115, line 41: Do we still want a verticalKey of `people` here?

sample-app/src/hooks/usePage.tsx
```diff
@@ -0,0 +1,33 @@
+import { SearchIntent } from "@yext/answers-headless";
+import { useLayoutEffect } from "react";
+import { useAnswersActions } from "@yext/answers-headless-react";
+import { executeSearch, getSearchIntents, updateLocationIfNeeded } from "../utils/search-operations";
+
+/**
+ * Sets up the state for a page
+ * @param verticalKey - The verticalKey associated with the page, or undefined for universal pages
+ */
+export default function usePage(verticalKey?: string) {
```
tmeyer2115, line 10: Nit: Could we call this `setupPage` or something like that? The `use` prefix suggests a hook which will return a value and a setter.

sample-app/src/hooks/usePage.tsx
```diff
@@ -0,0 +1,33 @@
...
+  const answersActions = useAnswersActions();
+  useLayoutEffect(() => {
+    const stateToClear = verticalKey
+      ? { universal: {} }
+      : { vertical: {} };
+    answersActions.setState({
```
yen-tt, line 16: should we add a reset on facets and filters for this page setup? I noticed sometimes query may change between pages, but navigating back to previous pages with filters/facets, the selected filters/facets may not correspond with the new query. Unless we want to leave this behavior up to user?

sample-app/src/hooks/usePage.tsx
yen-tt: I think this needs to have a use prefix to be consider a custom hook so it can call other hooks (e.g. useAnswersActions, useLayoutEffect), otherwise react would yell at you.

sample-app/src/App.tsx
cea2aj: I don't see a reason to keep it, so I'll remove it

sample-app/src/hooks/usePage.tsx
cea2aj: I like 'useAnswersOnPage'

sample-app/src/hooks/usePage.tsx
cea2aj: Ya, I noticed that as well. I'll update to clear filters here as well

sample-app/src/pages/LocationsPage.tsx
```diff
@@ -0,0 +1,72 @@
...
+  entityType: 'location'
+}, {
+  fieldApiName: 'services',
+  entityType: 'location'
+}];
+
+export default function LocationsPage({ verticalKey }: {
+  verticalKey: string
+}) {
+  useAnswersOnPage(verticalKey);
```
oshi97, line 27: `useAnswersOnPage` makes me think the hook is returning some `answers` variable

oshi97: maybe the name can say something about seeding initial state or setting up the page? I was thinking `useAnswersPageSetup`

oshi97: or even something like `usePageSetupEffect`? I know I'm bringing up an already resolved comment but I still think the name could be tweaked a little

cea2aj: Sure, i'll update to `usePageSetupEffect`

oshi97 (approved):

PR #85, opened 2021-12-08
add service-worker caching

oshi97:
This commit adds service worker caching for
universal search, universal autocomplete, and vertical autocomplete.
The VisualAutocomplete spec only called for caching of
universal search, however it's really simple to add additional caching
for the 2 autocompletes, and makes the experience feel very snappy.

Talked with Rose and we somewhat arbitrarily chose 30 seconds.
There should be no issues with caching anything autocomplete related
since we don't charge for autocomplete. However, universal searches
not made for VisualAutocomplete should not be cached.

J=SLAP-1729
TEST=manual

universal searches made for entity previews, and regular autocomplete
requests were both cached, by going into the network tab and seeing ~1-2ms response
times and the Size column saying "(ServiceWorker")

cache expiration works as expected, and will expire after 30 seconds

regular universal searches were never cached

there's a bug right now in VisualSearchBar where if you have no recent
searches, but hit the down arrow, the whole app crashes

### DIFF
--- a/sample-app/config/webpack.config.js
+++ b/sample-app/config/webpack.config.js
@@ -672,8 +672,12 @@ module.exports = function (webpackEnv) {
       new webpack.IgnorePlugin(/^\.\/locale$/, /moment$/),
       // Generate a service worker script that will precache, and keep up to date,
       // the HTML & assets that are part of the webpack build.
-      isEnvProduction &&
-        fs.existsSync(swSrc) &&
+      /**
+       * @remarks
+       * CRA by default does not build service workers in dev mode.
+       * We do, because we have custom logic for caching API requests.
+       */
+      fs.existsSync(swSrc) &&
         new WorkboxWebpackPlugin.InjectManifest({
           swSrc,
           dontCacheBustURLsMatching: /\.[0-9a-f]{8}\./,
--- a/sample-app/src/registerServiceWorker.ts
+++ b/sample-app/src/registerServiceWorker.ts
@@ -1,0 +1,6 @@
+export default function registerServiceWorker() {
+  window.addEventListener('load', () => {
+    const swUrl = `${process.env.PUBLIC_URL}/service-worker.js`;
+    navigator.serviceWorker.register(swUrl)
+  });
+}
--- a/sample-app/src/service-worker.ts
+++ b/sample-app/src/service-worker.ts
@@ -1,0 +1,71 @@
+/// <reference lib="webworker" />
+import { clientsClaim } from 'workbox-core';
+import { precacheAndRoute } from 'workbox-precaching';
+
+declare const self: ServiceWorkerGlobalScope;
+
+clientsClaim();
+const precacheManifest = self.__WB_MANIFEST;
+// Only precache static assets in production.
+// Precaching in development can result in confusing situations.
+// See https://github.com/facebook/create-react-app/issues/2398#issuecomment-304638935
+if (process.env.NODE_ENV !== 'development') {
+  precacheAndRoute(precacheManifest);
+}
+
+const cacheName = 'YEXT_ANSWERS_CACHE';
+const universalSearch = /v2\/accounts\/me\/answers\/query/;
+const universalAutocomplete = /v2\/accounts\/me\/answers\/autocomplete/;
+const verticalAutocomplete = /v2\/accounts\/me\/answers\/vertical\/autocomplete/;
+const swFetchedOnHeader = 'yxt-sw-fetched-on';
+
+// Clear the cache when a new service worker activates
+self.addEventListener('activate', event => {
+  event.waitUntil(caches.delete(cacheName));
+});
+
+self.addEventListener('fetch', (event: FetchEvent) => {
+  // Only cache universal search requests intended for VisualAutocomplete
+  if (
+    universalSearch.test(event.request.url) &&
+    new URLSearchParams(event.request.url).has('restrictVerticals')
+  ) {
+    event.respondWith(respondUsingCache(event.request, 30 * 1000));
+  }
+
+  if ([universalAutocomplete, verticalAutocomplete].find(regex => regex.test(event.request.url))) {
+    event.respondWith(respondUsingCache(event.request, 30 * 1000));
+  }
+});
+
+async function respondUsingCache(request: Request, timeToLiveInMilliseconds: number): Promise<Response> {
+  const cachedResponse = await caches.match(request);
+  const fetchedOn = cachedResponse?.headers.get(swFetchedOnHeader);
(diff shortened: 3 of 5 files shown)


sample-app/src/service-worker.ts
```diff
@@ -0,0 +1,71 @@
+/// <reference lib="webworker" />
+import { clientsClaim } from 'workbox-core';
+import { precacheAndRoute } from 'workbox-precaching';
+
+declare const self: ServiceWorkerGlobalScope;
```
tmeyer2115, line 5: Reading the ServiceWorker docs, I came across this statement: "After the activation step, the service worker will control all pages ... though the page that registered the service worker ... won't be controlled until it's loaded again." Do we not care about this because `registerServiceWorker` happens outside the scope of `App`?

sample-app/src/registerServiceWorker.ts
```diff
@@ -0,0 +1,6 @@
+export default function registerServiceWorker() {
+  window.addEventListener('load', () => {
+    const swUrl = `${process.env.PUBLIC_URL}/service-worker.js`;
```
tmeyer2115, line 3: Is `PUBLIC_URL` something `create-react-app` makes available for us automagically?

sample-app/src/service-worker.ts
```diff
@@ -0,0 +1,71 @@
...
+
+clientsClaim();
+const precacheManifest = self.__WB_MANIFEST;
+// Only precache static assets in production.
+// Precaching in development can result in confusing situations.
+// See https://github.com/facebook/create-react-app/issues/2398#issuecomment-304638935
+if (process.env.NODE_ENV !== 'development') {
+  precacheAndRoute(precacheManifest);
```
tmeyer2115, line 13: Out of curiosity, how much does pre-caching the manifest get us? Could it make the installation of the service worker more brittle than if we didn't cache any static assets?

sample-app/src/service-worker.ts
```diff
@@ -0,0 +1,71 @@
...
+  }
+
+  if ([universalAutocomplete, verticalAutocomplete].find(regex => regex.test(event.request.url))) {
+    event.respondWith(respondUsingCache(event.request, 30 * 1000));
+  }
+});
+
+async function respondUsingCache(request: Request, timeToLiveInMilliseconds: number): Promise<Response> {
+  const cachedResponse = await caches.match(request);
+  const fetchedOn = cachedResponse?.headers.get(swFetchedOnHeader);
```
tmeyer2115, line 43: Could we give this a more descriptive name? I'm not really sure what `fetchedOn` means

sample-app/src/service-worker.ts
```diff
@@ -0,0 +1,71 @@
...
+    }
+  }
+
+  const [cache, freshResponse] = await Promise.all<Cache, Response>([
+    caches.open(cacheName),
+    fetch(request)
+  ]);
+  const clone = freshResponse.clone();
+  const headers = new Headers(freshResponse.headers);
+  headers.append(swFetchedOnHeader, new Date().getTime().toString());
```
tmeyer2115, line 62: For my own edification, could you explain why we need to add this special header?

sample-app/src/registerServiceWorker.ts
oshi97: yup! this particular line is taken from the example service worker registration in the PWA (progressive web app) create-react-app template https://github.com/cra-template/pwa/blob/master/packages/cra-template-pwa-typescript/template/src/serviceWorkerRegistration.ts

sample-app/src/service-worker.ts
oshi97: Precaching static assets speeds up subsequent visits, and also allows for offline support. We get nearly instant load speeds for static js/css, which is kind of scope creep on this pr but seemed way too juicy to pass up. (I'm very excited about this)

image

Precaching doesn't affect the brittleness of the service worker, but it does introduce a little complexity on making sure clients always have the most up to date assets. I'm glad you asked me about this. I did some testing, and out of the box, updates to the service worker will only be activated once all pages using the old service worker have been closed, and the page would also have to be refreshed an additional time to use the updated precached assets. This aligns with what all of the service worker documentation I read on the web says.

To remedy this I added a `self.skipWaiting()` to the install step as well as an autorefresh to the service worker registration. This is based on what the create-react-app PWA template does. The autorefresh seems like a hack, but because of the precached assets the refresh occurs almost instantly.

sample-app/src/service-worker.ts
oshi97: renamed to `fetchedOnTimestampHeader` if that works?

sample-app/src/service-worker.ts
oshi97: We need to know when Responses in the cache have become stale.
The service worker's [cache](https://developer.mozilla.org/en-US/docs/Web/API/Cache) is a mapping of Request objects to Response objects. And there isn't any extra place (from what I can see, anyways) to tack on extra metadata except in the headers. I took this approach from [here](https://gomakethings.com/how-to-set-an-expiration-date-for-items-in-a-service-worker-cache/) and [here](https://phyks.me/2019/01/manage-expiration-of-cached-assets-with-service-worker-caching.html)

sample-app/src/service-worker.ts
oshi97: Yes! I thought using clientsClaim (i.e. `self.clients.claim()`) would take care of this, but that isn't exactly the case. To make this a more seamless experience I added some additional logic based on [this](https://github.com/cra-template/pwa/blob/25cb9f06938c671507684a8438c347d3c8bc8b24/packages/cra-template-pwa-typescript/template/src/serviceWorkerRegistration.ts)

sample-app/config/webpack.config.js
```diff
@@ -672,8 +672,12 @@ module.exports = function (webpackEnv) {
       new webpack.IgnorePlugin(/^\.\/locale$/, /moment$/),
       // Generate a service worker script that will precache, and keep up to date,
       // the HTML & assets that are part of the webpack build.
-      isEnvProduction &&
-        fs.existsSync(swSrc) &&
+      /**
+       * @remarks
+       * CRA by default does not build service workers in dev mode.
```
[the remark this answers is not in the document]
oshi97: CRA is create-react-app. I added it as `@remarks` to try to differentiate it between "comments we wrote" and "comments that came with this webpack config", but git history is a thing so maybe that's not necessary. (will also rename to create-react-app)

tmeyer2115 (approved):

cea2aj:
This is cool stuff

PR #87, opened 2021-12-09
Fire the correct search on verticals

yen-tt:
update searchbar to ensure execute right universal/vertical execution

J=none
TEST=manual

see that vertical search page now updates result correctly

### DIFF
--- a/sample-app/src/pages/StandardLayout.tsx
+++ b/sample-app/src/pages/StandardLayout.tsx
@@ -26,7 +26,7 @@ const StandardLayout: LayoutComponent = ({ page }) => {
       {isVertical
         ? <SearchBar
           placeholder='Search...'
-          isVertical={true}
+          isVertical={isVertical}
           screenReaderInstructionsId='SearchBar__srInstructions'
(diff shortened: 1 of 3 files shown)


cea2aj (approved):

cea2aj:
Can you update the PR title to something a little more descriptive such as "Fire the correct search on verticals"?

sample-app/src/pages/StandardLayout.tsx
```diff
@@ -26,7 +26,7 @@ const StandardLayout: LayoutComponent = ({ page }) => {
       {isVertical
         ? <SearchBar
           placeholder='Search...'
-          isVertical={true}
+          isVertical={isVertical}
```
oshi97, line 29: I was thinking originally that if this was java intellij would yell at me for using a variable for a value that was always true (here isVertical is always true since it's guarded by the ternary operator) but on second thought it's probably better to use the variable to make copy pasting less error prone

oshi97 (approved):

PR #88, opened 2021-12-10
Connected Searchbars Experience

yen-tt:
Changes:
- Used context to provide a shared state of search bar inputs so top-right search bar can manipulate the other search bar
- Updated SearchBar/VisualSearchBar to maintain a separate input value instead of directly use on query in headless state
- css styling
- new layout to have two search bars on page

Notes:
- input value in SearchBar/VisualSearchBar currently directly display query store in state. This makes it difficult to differentiate the input values between difference instances of SearchBar/VisualSearchBar (e.g. any keystroke that update query will update every searchbar instances). Instead, SearchBar/VisualSearchBar may have a separate input value maintain in component state that updates in tandem to executeQuery calls corresponding to that specific instances.
- In order for top-right search bar to update the other main-experience search bar under specific conditions (e.g. on executed search, and only one update the other and not vice versa), the search bar input must be able to be modify outside of its own instance. So, search bar would need to change so that it also consider input updates in a share context.

J=SLAP-1731
TEST=manual

- See that there's two searchbars in universal/vertical pages
- See that executed query in main-experience search bar does not affect the top-right search bar. See the corresponding results on page.
- See that executed query in top-right search bar updated the query in main-expeirence search bar. See the corresponding results loaded on page.

### DIFF
--- a/sample-app/src/App.tsx
+++ b/sample-app/src/App.tsx
@@ -2,11 +2,11 @@ import UniversalSearchPage from './pages/UniversalSearchPage';
 import FAQsPage from './pages/FAQsPage';
 import EventsPage from './pages/EventsPage';
 import PageRouter from './PageRouter';
-import StandardLayout from './pages/StandardLayout';
 import { AnswersHeadlessProvider } from '@yext/answers-headless-react';
 import { universalResultsConfig } from './universalResultsConfig';
 import JobsPage from './pages/JobsPage';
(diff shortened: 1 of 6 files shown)


yen-tt:
https://user-images.githubusercontent.com/36055303/145589609-175966ab-f289-488f-ab17-b28dcf644427.mov

PR #90, opened 2021-12-10
Remove Sample App

yen-tt:
- remove sample app folder
- remove wcag github action
- remove sample app from configs and package json

J=SLAP-1741
TEST=none


oshi97:
can we fork answers-headless-react and make the separate sample-app repo before this PR? I was thinking it could be nice to keep the git history, instead of having a giant single commit in the new repo

yen-tt:
> can we fork answers-headless-react and make the separate sample-app repo before this PR? I was thinking it could be nice to keep the git history, instead of having a giant single commit in the new repo

I see, yea that sounds good. I don't have permission to fork this to yext org, and I would need someone to delete the current new sample-app repo. Do you or @tmeyer2115 have the permission to do so?

oshi97:
I don't think you would need to delete the current repo, I believe that within the current react-sample-app, you can set the react-starter-site repo as a remote origin and force push to that

yen-tt:
got it. the main branch on the new repo is updated to match main branch here, with all the commits!

PR #91, opened 2021-12-13
version bump and update generate-notices

yen-tt:
- bump version to 0.4.3 (-beta is removed to follow similar syntax as headless repo)
- ~~update generate-notices to not include current repo's license~~ will include this in another pr after fixing an issue in the script

J=SLAP-1746
test=manual

- see that third-party-notices file are generated correctly


THIRD-PARTY-NOTICES
```diff
@@ -60,13 +60,52 @@ SOFTWARE.
 
 -----------
 
-The following NPM packages may be included in this product:
+The following NPM package may be included in this product:
 
  - @yext/answers-core@1.5.0-beta.1
- - @yext/answers-headless-react@0.4.2-beta.0
- - @yext/answers-headless@0.2.0
```
oshi97, line 67: I don't think answers-headless should be wiped here?

yen-tt: answers-headless is included in the following package license right after this one

oshi97: Ah ok cool! That's probably because of the license updates in 0.3.0? I still think the remove-repo-license script needs to be revisited, maybe it would make sense to not include it for now?

yen-tt: yeah I think it's because if the 0.3.0 update. You did raise a good about the script issue though, I will look into it after finishing up this facet update item. I can remove the script for now but it would include the repo's own license in third party licenses, which may be a little weird. @tmeyer2115 had a question on this in another pr, and this script was the approach we have in answers-core to prevent that so I just modified the script a bit for it to work outside of core. I'll leave it out of this pr, and l try to see if I can resolve this before we publish v0.9

.eslintrc.js
```diff
@@ -6,7 +6,7 @@ module.exports = {
     'plugin:@yext/export-star/recommended'
   ],
   plugins: ['react-perf'],
-  ignorePatterns: ['lib', 'tests/setup/responses'],
+  ignorePatterns: ['lib', 'tests/setup/responses', 'scripts'],
```
oshi97, line 9: oh we can get rid of scripts here for now

yen-tt: yea, my b

PR #110, opened 2022-03-18
Pass Client-SDK agent to Headless

nmanu1:
Passes the `ANSWERS_HEADLESS_REACT` agent as part of the Client-SDK header to Headless through `provideAnswersHeadless`.

J=SLAP-1951
TEST=manual

Hook up locally to the starter app and see that requests send the Headless React, Headless, and Core agents as part of the Client-SDK header. Requests to the universal search endpoint made for entity previews only send the Headless and Core agents because the hook uses `provideAnswersHeadless` directly to get an instance of Headless instead of going through the `AnswersHeadlessProvider` component from Headless React.

### DIFF
--- a/THIRD-PARTY-NOTICES
+++ b/THIRD-PARTY-NOTICES
@@ -62,7 +62,7 @@ SOFTWARE.
 
 The following NPM package may be included in this product:
 
- - @yext/answers-core@1.6.0-beta.3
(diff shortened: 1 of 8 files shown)


cea2aj (approved):
LGTM

oshi97 (approved):
PogChamp

PR #111, opened 2022-03-24
use shared eslint and fix eslint issues

oshi97:
use the new slapshot shared eslint
fixes issues, including non autofixable ones
the non autofixble ones were
- conditionally callint jest.expect
- wrapping react-testing-library calls inside an act() unnecessarily
- tsdoc params expect a hyphen

J=SLAP-2003
TEST=build


cea2aj (approved):

oshi97:
force pushed back to original commit this was approved

PR #117, opened 2022-04-25
Isomorphic Layout fix and JSON type assertion

cea2aj:


src/utils/useIsomorphicLayoutEffect.ts
```diff
@@ -0,0 +1,3 @@
+import { useEffect, useLayoutEffect } from 'react';
+
+export default typeof document !== 'undefined' ? useLayoutEffect : useEffect;
```
oshi97, line 3: what were the issues with using the package directly? the code seeems to be the same? is this to get local npm links working?

src/AnswersHeadlessProvider.tsx
```diff
@@ -2,7 +2,7 @@ import { ReactChild, ReactChildren } from 'react';
 import { provideAnswersHeadless, AnswersHeadless, HeadlessConfig } from '@yext/answers-headless';
 import { AnswersHeadlessContext } from './AnswersHeadlessContext';
 import acquireSessionId from './utils/acquireSessionId';
-import packageJson from '../package.json';
+import packageJson from '../package.json' assert { type: 'json' };
```
oshi97, line 5: didn't know about [import assertions](https://v8.dev/features/import-assertions)!
do we also need to update our other repos that import json files?

src/utils/useIsomorphicLayoutEffect.ts
cea2aj: In our local testing, it was trying to load the CJS build which was causing some issues. We felt it was simplest to just use our own utility here

src/AnswersHeadlessProvider.tsx
cea2aj: Tom will talk to Max about them since older versions of node blow up when they encounter them. Support was added in node 17.1 and was made required in node 17.5 according to a [stackoverflow post](https://stackoverflow.com/questions/70106880/err-import-assertion-type-missing-for-import-of-json-file). It was also added to 16.14 however Tom wanted us to simply require node 17.1 for now and publish an alpha for testing purposes

src/utils/useIsomorphicLayoutEffect.ts
oshi97: Is there some shared place we can put this? I would rather not copy paste this into multiple repos even though it's a one liner. Is this the only dependency that has this issue? It might be worth spending more time investigating the build infra if this means we can't use third party deps

cea2aj: I suspect the issue stems from the fact that the yext sites infra uses node's experimental ESM imports rather than CJS. The useIsomorphicLayoutEffect library defines the ESM with the 'module' property, however that is ignored by node. I expect the issue would be fixed if the library added an 'exports' prop to the package.json file which defines where the ESM module is imported from

src/AnswersHeadlessProvider.tsx
cea2aj: The import assertion is causing me some issues since Vite doesn't support it, so we can leave it out for now

oshi97 (approved):

src/utils/useIsomorphicLayoutEffect.ts
oshi97: What is 'exports'? It sounds familiar but I couldn't find it anywhere in the npm docs here https://docs.npmjs.com/cli/v8/configuring-npm/package-json for v6, 7, or 8

oshi97: Ah, I see it here https://nodejs.org/api/packages.html#exports

cea2aj: The exports property supports [conditional exports](https://nodejs.org/api/packages.html#conditional-exports) which allows node to use either ESM or CJS. It also allows subpaths to be defined which is what we use in the answers-search-ui package.json file for the i18n builds

PR #118, opened 2022-05-02
add commonjs support, remove type: module

oshi97:
This PR adds commonjs support by separating the tsconfig into esm and commonjs versions,
updating our api extractor config, and using conditional exports to provide the right entry point.

At one point, this PR added commonjs support by instead building the library with module: commonjs,
and adding a thin esm wrapper to also support es6 syntax.
This was in order to prevent the dual package hazard documented in
https://nodejs.org/api/packages.html#dual-commonjses-module-packages
and
https://github.com/GeoffreyBooth/dual-package-hazard/blob/master/index.mjs
However, from personal testing vite was not able to tree shake our esm-wrapper.

Furthermore, the dual package hazard is primarily a concern when different plugin-like libraries
are written around some central library, for example if people were to write extensions of
answers-headless-react, and these extensions were written in a mix of only esm and only cjs.
Then, users who tried to use both plugins would find themselves with the dual package hazard.
This scenario is very unlikely because we can expect the vast majority of new code to be written in esm.
As a general trend ESM is the new standard for JS libraries, and is also promoted by typescript.

J=SLAP-2074
TEST=manual

test that the dev command works - I can hook up a local answers-headless-react to a local
project, and changes in answers-headless-react will automatically be reflected in my project

test that answers-headless-react works with yext sites SSR in both prod and dev mode
this was tested using npm pack


cea2aj:
nit: can you update the PR title since this doesn't use an esm wrapper?

cea2aj (approved):

oshi97:
> nit: can you update the PR title since this doesn't use an esm wrapper?

updated!

PR #122, opened 2022-05-23
support React 18

yen-tt:
This pr updates answers-headless-react to support React 18 as well as 17.

- use alpha version 1.1.1.-alpha.0-95 of answers-headless which contains redux package update to support React 18.
- to support react 18 concurrent mode, useSyncExternalStore is needed. This pr uses `use-sync-external-store` package for backwards-compatible with pre-18 versions (v17).
  - ~~From [internal doc](https://github.com/facebook/react/blob/main/packages/use-sync-external-store/src/useSyncExternalStoreShimServer.js#L15)s in `use-sync-external-store`, the package does not offer much logic to support the case of server side rendering for react pre-18 versions. So, this pr update useAnswersState to be conditionally exported with either `useAnswersStateWithManualStoreSync` for SSR pre-18 or `useAnswersStateWithReactStoreSync` for everything else.~~ updated to use `use-sync-external-store` as is, without the manual handling, for all cases ([for more info](https://github.com/yext/answers-headless-react/pull/122#issuecomment-1136456564)).
  - with v17 version, there's warnings for state update on unmounted components, which we do get for when an async function (execute universal/vertical search) finish and send a series of dispatch that would trigger a re-render of a component that's already unmounted. There's no actual memory leak here since we always unsubscribe out listeners. This warning was removed in v18 (visit this [PR](https://github.com/facebook/react/pull/22114) for more details). As a workaround, `isMountedRef` is used in `useAnswersStateWithReactStoreSync` to track when a component is unmounted and prevent the callback for handlling store changes from triggering (cause component update).
  - update relevant packages to support v18 as well (`@testing-library/react`, `@types/react`)

Note: will release an alpha version of this for the component lib.

J=SLAP-2095
TEST=manual&auto
- added new jest tests for server side rendering and hydration process.
- see that jest tests passed successfully for V17 and V18. (note that @testing-library/react v13 drop support for 17 when beginning to support 18 in the same version)
- test (client side) with React version 17 and React version 18 in test-site from component lib repo.
  - with React verison 18. test with render (pre-18 syntax) and createRoot (18), ran a couple search and switch between universal and vertical page. see that both works as expected, although there is a warning for the former approach (`Warning: ReactDOM.render is no longer supported in React 18. Use createRoot instead. Until you switch to the new API, your app will behave as if it's running React 17.`) which is expected.
- test (SSR) with React version 17 and React version 18 with yext-site-starters repo. See that the assets were built successfully and the page function as expected.

### DIFF
--- a/package.json
+++ b/package.json
@@ -22,7 +22,6 @@
   },
   "dependencies": {
     "@yext/answers-headless": "^1.1.0-alpha.0-95",
-    "use-isomorphic-layout-effect": "^1.1.2",
     "use-sync-external-store": "^1.1.0"
   },
   "devDependencies": {
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@yext/answers-headless-react",
-  "version": "1.1.0-beta.13",
+  "version": "1.1.1-alpha.0",
   "main": "./lib/esm/src/index.js",
   "types": "./lib/esm/src/index.d.ts",
   "exports": {
@@ -21,8 +21,9 @@
     "generate-notices": "generate-license-file --input package.json --output THIRD-PARTY-NOTICES --overwrite"
   },
   "dependencies": {
-    "@yext/answers-headless": "^1.1.0-beta.12",
-    "use-isomorphic-layout-effect": "^1.1.2"
+    "@yext/answers-headless": "^1.1.0-alpha.0-95",
+    "use-isomorphic-layout-effect": "^1.1.2",
+    "use-sync-external-store": "^1.1.0"
   },
   "devDependencies": {
     "@babel/preset-env": "^7.16.5",
@@ -30,10 +31,11 @@
     "@babel/preset-typescript": "^7.16.5",
     "@css-modules-theme/core": "^2.3.0",
     "@testing-library/jest-dom": "^5.16.1",
-    "@testing-library/react": "^12.1.2",
+    "@testing-library/react": "^13.2.0",
     "@testing-library/user-event": "^13.5.0",
     "@types/jest": "^27.0.3",
-    "@types/react": "^17.0.37",
+    "@types/react": "^18.0.9",
+    "@types/use-sync-external-store": "^0.0.3",
     "@types/uuid": "^8.3.3",
     "@typescript-eslint/eslint-plugin": "^5.16.0",
     "@typescript-eslint/parser": "^5.16.0",
@@ -51,14 +53,14 @@
     "generate-license-file": "^1.3.0",
     "jest": "^27.4.5",
     "msw": "^0.36.3",
-    "react": "^17.0.2",
-    "react-dom": "^17.0.2",
+    "react": "^18.1.0",
+    "react-dom": "^18.1.0",
     "ts-jest": "^27.1.1",
     "typescript": "^4.5.4",
     "uuid": "^8.3.2"
   },
   "peerDependencies": {
-    "react": "^17.0.2"
+    "react": "^17.0.0 || ^18.0.0"
   },
   "jest": {
     "bail": 0,
@@ -82,7 +84,7 @@
     ],
     "testEnvironment": "jsdom",
     "testMatch": [
-      "<rootDir>/tests/**/*.(test).ts(x)?"
+      "<rootDir>/tests/useAnswersState.test.tsx"
     ],
     "globals": {
       "ts-jest": {
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@yext/answers-headless-react",
-  "version": "1.1.1-alpha.0-122",
+  "version": "1.1.1-alpha-122",
   "main": "./lib/esm/src/index.js",
   "types": "./lib/esm/src/index.d.ts",
   "exports": {
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@yext/answers-headless-react",
-  "version": "1.1.1-alpha.0",
+  "version": "1.1.1-alpha.0-122",
   "main": "./lib/esm/src/index.js",
   "types": "./lib/esm/src/index.d.ts",
   "exports": {
@@ -84,7 +84,7 @@
     ],
     "testEnvironment": "jsdom",
     "testMatch": [
-      "<rootDir>/tests/useAnswersState.test.tsx"
+      "<rootDir>/tests/**/*.(test).ts(x)?"
     ],
     "globals": {
       "ts-jest": {
--- a/package.json
+++ b/package.json
@@ -59,7 +59,7 @@
     "uuid": "^8.3.2"
   },
   "peerDependencies": {
-    "react": "^17.0.0 || ^18.0.0"
+    "react": "^17 || ^18"
   },
   "jest": {
     "bail": 0,
--- a/src/AnswersHeadlessProvider.tsx
+++ b/src/AnswersHeadlessProvider.tsx
@@ -1,4 +1,4 @@
-import { ReactChild, ReactChildren } from 'react';
+import { PropsWithChildren } from 'react';
 import { provideAnswersHeadless, AnswersHeadless, HeadlessConfig } from '@yext/answers-headless';
 import { AnswersHeadlessContext } from './AnswersHeadlessContext';
 import acquireSessionId from './utils/acquireSessionId';
@@ -7,12 +7,11 @@ import packageJson from '../package.json';
 const { version } = packageJson;
 
 type Props = HeadlessConfig & {
-  children?: ReactChildren | ReactChild | (ReactChildren | ReactChild)[],
   verticalKey?: string,
   sessionTrackingEnabled?: boolean
 };
 
-export function AnswersHeadlessProvider(props: Props): JSX.Element {
+export function AnswersHeadlessProvider(props: PropsWithChildren<Props>): JSX.Element {
   const { children, verticalKey, sessionTrackingEnabled=true, ...answersConfig } = props;
   const additionalHttpHeaders = {
     'Client-SDK': {
--- a/src/useAnswersState.tsx
+++ b/src/useAnswersState.tsx
@@ -53,8 +53,7 @@ function useAnswersStateWithReactStoreSync<T>(stateSelector: StateSelector<T>): 
         }
         cb();
       }
-    })
-  , [answers]);
+    }), [answers]);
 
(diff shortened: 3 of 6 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@yext/answers-headless-react",
-  "version": "1.1.0-beta.13",
+  "version": "1.1.1-alpha.0-122",
```
oshi97, line 3: do we need the `.0` part of the tag? I feel like we could just do `1.1.1-alpha-122` or some variant of that

src/useAnswersState.tsx
```diff
@@ -1,15 +1,76 @@
...
+
+/**
+ * use-sync-external-store/shim does not support getServerSnapshot for pre-18 React versions.
+ * It requires user of the shim to handle subscription logic and return the correct snapshot value.
+ * As such, server side rendering from pre-18 React version will use a version of useAnswersState that
+ * manually handle memoization of state selector and register subscription without useSyncExternalStore call.
+ */
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export const useAnswersState = (React as any).useSyncExternalStore === undefined && isServerEnvironment
+  ? useAnswersStateWithManualStoreSync : useAnswersStateWithReactStoreSync;
```
oshi97, line 22: so `useAnswersStateWithManualStoreSync` is only for react 17 SSR?

oshi97: does `useAnswersStateWithManualStoreSync` work with react 18 as well?

src/useAnswersState.tsx
```diff
@@ -1,15 +1,76 @@
...
+      valueAccessor: state => state,
+      callback: () => {
+        // prevent React state update on an unmounted component
+        if (!isMountedRef.current) {
+          return;
+        }
+        cb();
+      }
+    })
+  , [answers]);
```
oshi97, line 57: would it look better with the comma on the previous line instead?

src/useAnswersState.tsx
oshi97: I'm also kind of curious if you read anything about what useSyncExternalStore is doing under the hood? If not then nbd

src/AnswersHeadlessProvider.tsx
```diff
@@ -7,12 +7,11 @@ import packageJson from '../package.json';
 const { version } = packageJson;
 
 type Props = HeadlessConfig & {
-  children?: ReactChildren | ReactChild | (ReactChildren | ReactChild)[],
```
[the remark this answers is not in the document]
yen-tt: 'ReactChildren' and `ReactChild` is marked as deprecated. There's also a typescript error `Type 'ReactChildren' is not assignable to type 'ReactNode'.` when upgrade to v18 (works fine for v17). Switching to `PropsWithChildren` seems to work well here

src/useAnswersState.tsx
yen-tt: When I was testing with `startTransition` to try to simulate tearing issues, `useAnswersStateWithManualStoreSync` doesn't work well with concurrent feature in react 18. 
Yup! `useAnswersStateWithManualStoreSync` would be for react 17 SSR only since useSyncExternalStore can handle the client side rendering for 17 and automatically use the builtin useSyncExternalStore API for 18. 
I didn't dig too much into what useSyncExternalStore does internally but it looks like it is a [reform](https://github.com/reactwg/react-18/discussions/86) of [useMutableSource](https://github.com/reactwg/react-18/discussions/84) if you are familiar with that

src/useAnswersState.tsx
```diff
@@ -1,15 +1,75 @@
...
+export const useAnswersState = (React as any).useSyncExternalStore === undefined && isServerEnvironment
+  ? useAnswersStateWithManualStoreSync : useAnswersStateWithReactStoreSync;
+
+
+/**
+ * Returns the Answers State returned by the map function.
+ * UseSyncExternalStoreWithSelector is used to subscribe to external store and trigger rerender
+ * whenever an action is dispatched and caused changes to the state of the selector function.
+ */
+function useAnswersStateWithReactStoreSync<T>(stateSelector: StateSelector<T>): T {
```
oshi97, line 30: would it make sense to separate these two functions into separate files?

src/useAnswersState.tsx
```diff
@@ -1,15 +1,75 @@
...
 
 export type StateSelector<T> = (s: State) => T;
 
+const canUseDOM = !!(typeof window !== 'undefined'
+  && typeof window.document !== 'undefined'
+  && typeof window.document.createElement !== 'undefined');
+const isServerEnvironment = !canUseDOM;
+
+/**
+ * use-sync-external-store/shim does not support getServerSnapshot for pre-18 React versions.
```
oshi97, line 15: Could we explain what getServerSnapshot is, or write this comment in a way where people don't have to know?

oshi97:
Have we tried looking at what react-redux is doing? It looks like we match up pretty closely? https://github.com/reduxjs/react-redux/blob/master/src/hooks/useSelector.ts

src/useAnswersState.tsx
```diff
@@ -1,15 +1,75 @@
...
+      callback: () => {
+        // prevent React state update on an unmounted component
+        if (!isMountedRef.current) {
+          return;
+        }
+        cb();
+      }
+    }), [answers]);
+
+  const selectedState = useSyncExternalStoreWithSelector(
```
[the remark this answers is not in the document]
yen-tt: sure thing

yen-tt:
> Have we tried looking at what react-redux is doing? It looks like we match up pretty closely? https://github.com/reduxjs/react-redux/blob/master/src/hooks/useSelector.ts It looks like they support react 16 and 17 though so I'm kind of curious if you have any idea how they're doing so? I haven't looked suuuper closely but it looks like they're also doing some kind of conditional export with `src/index.ts` and `src/next.ts` https://github.com/reduxjs/react-redux/blob/master/package.json

After some digging into SSR, it look like component will NOT re-render (s) so the logic to handle memoization of state selector and register subscription (`useAnswersStateWithManualSyncStore`) is not necessary - useEffect, which is used in useAnswersStateWithManualSyncStore, [will not run in server](https://codewithhugo.com/react-useeffect-ssr/). This make sense with the way that `use-sync-external-store` and `react-redux` also doesn't do much with this case of SSR pre18 version other than directly returning the state snapshot. This would mean we need a way to apply initial state to redux BEFORE the call `renderToString` of the component in ssr, which would require some changes to `AnswersHeadlessProvider` and `configureStore` in answers-headless to accept a `preloadedState` option.

src/useAnswersState.tsx
yen-tt: removed `useAnswersStateWithManualSyncStore` so this is no longer needed

src/useAnswersState.tsx
yen-tt: removed `useAnswersStateWithManualSyncStore` so this change is no longer needed

package.json
```diff
@@ -51,14 +52,14 @@
...
-    "react-dom": "^17.0.2",
+    "react": "^18.1.0",
+    "react-dom": "^18.1.0",
     "ts-jest": "^27.1.1",
     "typescript": "^4.5.4",
     "uuid": "^8.3.2"
   },
   "peerDependencies": {
-    "react": "^17.0.2"
+    "react": "^17.0.0 || ^18.0.0"
```
oshi97, line 62: I think we can chop off the decimals and this can just be `^17 || ^18`

tests/useAnswersState.test.tsx
```diff
@@ -17,6 +18,37 @@ it('invoke useAnswersState outside of AnswersHeadlessProvider', () => {
   jest.clearAllMocks();
 });
 
+it('Retrieve state snapshot in server side rendering and hydration process', () => {
```
oshi97, line 21: maybe retrieve -> retrieves and in -> during ?

tests/useAnswersState.test.tsx
```diff
@@ -69,15 +101,15 @@ it('does not perform extra renders/listener registrations for nested components'
   expect(childStateUpdates).toHaveLength(0);
 
   userEvent.click(screen.getByText('Search'));
-  await pendingVerticalQuery;
+  await act( async () => { await pendingVerticalQuery; });
```
oshi97, line 104: out of curiosity why are these acts necessary?

src/useAnswersState.tsx
```diff
@@ -16,57 +17,32 @@ export function useAnswersState<T>(stateSelector: StateSelector<T>): T {
...
-    return storedSelectedState.current;
-  })();
-
-  const [, triggerRender] = useState<T>(storedSelectedState.current);
-  useLayoutEffect(() => {
-    storedSelector.current = stateSelector;
-    storedStoreState.current = answers.state;
-    storedSelectedState.current = selectedStateToReturn;
-  });
+  const getSnapshot = useCallback(() => answers.state, [answers.state]);
```
oshi97, line 20: What happens if this isn't wrapped in useCallback? totally fine if it's unknown don't need to deepdive just asking if you already knew. I guess I'm more concerned with if wrapping this in useCallback is the recommended usage?

src/useAnswersState.tsx
```diff
@@ -16,57 +17,32 @@ export function useAnswersState<T>(stateSelector: StateSelector<T>): T {
...
+    isMountedRef.current = true;
+    return () => {
+      isMountedRef.current = false;
+    };
+  }, []);
 
-  useLayoutEffect(() => {
-    let unsubscribed = false;
-    const unsubscribe = answers.addListener({
+  const subscribe = useCallback(cb =>
```
oshi97, line 29: same questions about useCallback here

src/useAnswersState.tsx
yen-tt: subscribe function and getSnapshot function are in the dependency array in the useLayoutEffect and useEffect calls in use-sync-external-store so having new function created every time would cause additional updates/renders if it's not wrapped in useCallback!

src/useAnswersState.tsx
yen-tt: ^ responded in the comment above

tests/useAnswersState.test.tsx
oshi97: Oh ok, maybe this is due to changing from react 17 to 18 with how batching works or something. I think I wrote this original test in a pretty weird way but I couldn't think of an easy way to change it here so eh

oshi97 (approved):

PR #124, opened 2022-05-26
add react 16.14 support to package.json + test different react versions

oshi97:
We support react 16.14 at latest - we do not support earlier versions of 16.
This is due to us using the newest jsx transform which was only added to 16.14 and not earlier versions of 16.
https://reactjs.org/blog/2020/09/22/introducing-the-new-jsx-transform.html

We could make answers-headless-react support earlier versions of react by changing our jsx transform
However in [yext/answers-react-components](https://github.com/yext/answers-react-components) we also depend on restart/ui which requires 16.14. Lastly, we already knew that redux-toolkit requires 16.9, which is another barrier
to support 16.8.

J=SLAP-2113
TEST=manual,auto

hook up local headless-react to local answers-react-components test site using react 16.14
see the components work as expected, and the dev tools say that the components are rendered by react-dom@16.14.0
try running vertical + universal searches, using numerical facets, using hierarchical facets

### DIFF
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -21,6 +21,6 @@ jobs:
     uses: yext/slapshot-reusable-workflows/.github/workflows/run_tests.yml@v1
     with:
       build_script: |
-        npm i -D react@16 react-dom@16 @testing-library/react@12
+        npm i -D react@16.14 react-dom@16.14 @testing-library/react@12
         npm run build
       node_matrix: '["16.x"]'
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -12,9 +12,12 @@ jobs:
       build_script: |
         npm i -D react@17 react-dom@17
         npm run build
+      node_matrix: '["16.x"]'
 
   call_run_tests:
     uses: yext/slapshot-reusable-workflows/.github/workflows/run_tests.yml@v1
+    with:
+      node_matrix: '["16.x"]'
 
   call_run_tests-react-16:
     uses: yext/slapshot-reusable-workflows/.github/workflows/run_tests.yml@v1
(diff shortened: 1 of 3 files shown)


.github/workflows/run-tests.yml
```diff
@@ -1,10 +1,26 @@
...
+        npm run build
+      node_matrix: '["16.x"]'
+
+  call_run_tests-react-16:
+    uses: yext/slapshot-reusable-workflows/.github/workflows/run_tests.yml@v1
+    with:
+      build_script: |
+        npm i -D react@16.14 react-dom@16.14 @testing-library/react@12
+        npm run build
+      node_matrix: '["16.x"]'
```
nmanu1, line 26: I thought we still support node versions older than 16?

.github/workflows/run-tests.yml
```diff
@@ -1,10 +1,26 @@
...
 on: [push, pull_request]
 
 jobs:
   call_run_tests:
     uses: yext/slapshot-reusable-workflows/.github/workflows/run_tests.yml@v1
+    with:
+      node_matrix: '["16.x"]'
+
+  # It's necessary to use v12 of the react-testing-library since v13 was updated to only support react 18
+  call_run_tests-react-17:
```
yen-tt, line 10: I made an item for running tests in React v17, potentially looking at `legacyRoot` for rendering in pre18 versions. but since we need to change the react version install, directly `@testing-library/react@12` make sense. Nice!

.github/workflows/run-tests.yml
oshi97: It felt a little egregious to run the unit tests 2 * 3 * 3 = 18 times. The first 2 is for on push/pull_request, the 3 is for 3 node versions, and the last 3 is for 3 react versions. We could take some kind of middle ground where only the react 18 tests are run for 3 node versions, but I think for headless-react it's probably fine to run the tests 18 times since the library is so small and probably will never need to do many other kinds of checks anyways.

tldr the tests run 18 times on prs now!

PR #127, opened 2022-06-07
react 16.8 support branch

oshi97:
Fixes the react and react-dom versions to 16.8
Changes the tsconfig react transform to React.createElement calls, and adds React imports to all
files since that is a requirement for the old jsx transform.

Update unit tests to pass, they were complaining that we shouldn't be awaiting an "act" call and
the tests were failing. Unfortunately async "act" calls were introduced in react 16.9 so some tests had
to be changed. One test had to be deleted, since it is now impossible to test in the same way and would
need to be completely rewritten. Since this is a one-off build, and react 16.8 users probably aren't knashing
their teeth over extra renders (which we shouldn't have any issues with unless we make a variety of large changes)
I just removed the test.

In useReducer, an argument is always required in react 16.8, even if you manually specify
the generic, so I passed it null.

J=SLAP-2129
TEST=manual

use create-react-app to create a react 16.8 test site
install my local answers-headless-react using `npm install ~/answers-headless-react --install-links`

Add a react 16.8 test site with simple answers components


react-16.8-test-site/.env
```diff
@@ -0,0 +1 @@
+SKIP_PREFLIGHT_CHECK=true
```
yen-tt, line 1: why do we need this? this would suppress any errors with the project dependency tree right?

react-16.8-test-site/package.json
```diff
@@ -0,0 +1,35 @@
+{
+  "name": "react16",
+  "version": "0.1.0",
+  "private": true,
+  "dependencies": {
+    "@yext/answers-headless-react": "file:..",
+    "react": "16.8.0",
+    "react-dom": "16.8.0",
```
yen-tt, line 8: I think you have to do:
```
"react": "file:../node_modules/react",
"react-dom": "file:../node_modules/react-dom",
```
if you are linking answers-headless-react with "file:..", otherwise you get two copies of React and it may get an error like `Uncaught Invariant Violation: Invalid hook call.  blah blah. You might have more than one copy of React in the same app`

tests/useAnswersState.test.tsx
```diff
@@ -136,9 +68,9 @@ it('does not trigger render on unmounted component', async () => {
     </AnswersHeadlessContext.Provider>
   );
   act(() => answers.setQuery('resultsWithFilter'));
-  await act( async () => { await answers.executeUniversalQuery(); });
+  await answers.executeUniversalQuery();
```
yen-tt, line 71: I believe `act()` is to help perform updates so stuff is properly re-render in DOM before you make any assertions. Without it, I'm not too sure just doing `await answers.executeUniversalQuery();` would trigger any updates for the test components to properly test the case of triggering on unmounted component

react-16.8-test-site/package.json
oshi97: I was relying on `--install-links` which avoids that issue, but I can update

tests/useAnswersState.test.tsx
oshi97: That's fair. I can delete the test

react-16.8-test-site/.env
oshi97: Yeah, it's not happy about the newer version of babel higher up in the tree. I didn't really see a good way around this, we technically don't need the test-site I just thought it might be easier for other people

react-16.8-test-site/src/App.js
```diff
@@ -0,0 +1,61 @@
...
+          Edit <code>src/App.js</code> and save to reload.
+        </p>
+        <a
+          className="App-link"
+          href="https://reactjs.org"
+          target="_blank"
+          rel="noopener noreferrer"
+        >
+          Learn React
+        </a>
```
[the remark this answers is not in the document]
oshi97: the spinning logo made me feel warm and fuzzy but I removed it all

react-16.8-test-site/.env
yen-tt: oh hmm, I wonder if doing `overrides` for babel version would help in test site. If not, this sounds fine!

react-16.8-test-site/package.json
yen-tt: gotcha, yea this might just be easier if we have the package linking in package.json. Some people might forget to do --install-links and that would cause confusion

PR #136, opened 2022-06-29
add reusable workflow for ESlint to answers-headless-react

juliannzhou:
This PR adds reusable workflow for ESlint to answers-headless-react. 

J=SLAP-2199
TEST=auto

### DIFF
--- a/.github/workflows/run-linting.yml
+++ b/.github/workflows/run-linting.yml
@@ -1,0 +1,10 @@
+name: Run Linter
+
+on:
+  pull_request:
+    branches:
+      - main
+
+jobs:
+  linting:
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -17,6 +17,7 @@
     "build:cjs": "tsc -p tsconfig.cjs.json",
     "prepublishOnly": "npm run build",
     "dev": "tsc --watch -p tsconfig.esm.json",
+    "lint": "eslint .",
     "test": "eslint . && jest",
```
oshi97, line 21: we can probably remove eslint from the test command since it's a separate workflow now

oshi97: also fine to keep but I'd rather not run the linting twice 🤷

yen-tt (approved):

PR #138, opened 2022-06-29
add description and keywords to package.json

juliannzhou:
This PR adds description and keywords to package.json. The description is copied from README and keywords provided by product.

J=SLAP-2190
TEST=manual

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,17 +1,15 @@
 {
   "name": "@yext/answers-headless-react",
   "version": "1.2.0",
-  "description": "Answers Headless React is the official React UI Bindings layer for Answers Headless",
+  "description": "The official React UI Bindings layer for Answers Headless",
   "main": "./lib/esm/src/index.js",
   "types": "./lib/esm/src/index.d.ts",
   "keywords": [
-    "search",
-    "nlp",
-    "ai",
-    "answers",
-    "yext",
-    "react",
-    "components"
+    "Search",
+    "React",
+    "Hooks",
+    "React Binding",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -17,6 +27,7 @@
     "build:cjs": "tsc -p tsconfig.cjs.json",
     "prepublishOnly": "npm run build",
     "dev": "tsc --watch -p tsconfig.esm.json",
+    "lint": "eslint .",
```
oshi97, line 30: is this from a separate PR?

package.json
```diff
@@ -1,8 +1,18 @@
...
   "main": "./lib/esm/src/index.js",
   "types": "./lib/esm/src/index.d.ts",
+  "keywords": [
+    "search",
+    "nlp",
+    "ai",
+    "answers",
+    "yext",
+    "react",
+    "components"
```
nmanu1, line 14: did product want "components" as a keyword here even though this library doesn't provide any components?

oshi97: This library technically has a component (AnswersHeadlessProvider) though it's really not about components

oshi97:
let's merge into develop for this one too

package.json
juliannzhou: Product only gave me one list, so I updated the repos with the same list. Should I ask product to give a list for every public repository?

nmanu1: Yeah, I would ask for a list for each repo since they have different purposes

oshi97 (approved):

PR #142, opened 2022-07-06
deprecate subscribeToStateUpdates

yen-tt:
deprecate `subscribeToStateUpdates`. Update README to provide an example for class component to use `AnswersHeadlessContext` directly to dispatch actions and receive updates from state.

J=SLAP-2208
TEST=manual

Test that the new Searcher example with `AnswersHeadlessContext` works as expected. See that README look as expected.

### DIFF
--- a/src/subscribeToStateUpdates.tsx
+++ b/src/subscribeToStateUpdates.tsx
@@ -7,12 +7,12 @@ import isShallowEqual from './utils/isShallowEqual';
 type SubscriberGenerator = (WrappedComponent: ComponentType<any>) => (props: any) => JSX.Element;
 
 /**
- * @deprecated
- * For class component, use `AnswersHeadlessContext` directly to dispatch actions and receive state updates.
- * For functional component, use `useAnswersAcions` and `useAnswersState` instead.
- *
  * Generates a HOC that updates a given Component's props based on the current
  * answers-headless state and a given mapping function.
+ *
+ * @deprecated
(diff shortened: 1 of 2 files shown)


oshi97 (approved):
nice!

src/subscribeToStateUpdates.tsx
```diff
@@ -11,6 +7,10 @@ import isShallowEqual from './utils/isShallowEqual';
 type SubscriberGenerator = (WrappedComponent: ComponentType<any>) => (props: any) => JSX.Element;
 
 /**
+ * @deprecated
+ * For class component, use `AnswersHeadlessContext` directly to dispatch actions and receive state updates.
+ * For functional component, use `useAnswersAcions` and `useAnswersState` instead.
```
nmanu1, line 12: nit: typo in `useAnswersActions`

src/subscribeToStateUpdates.tsx
```diff
@@ -11,6 +7,10 @@ import isShallowEqual from './utils/isShallowEqual';
 type SubscriberGenerator = (WrappedComponent: ComponentType<any>) => (props: any) => JSX.Element;
 
 /**
+ * @deprecated
```
nmanu1, line 10: can the `deprecated` tag and message be moved to the end? I think the docs get formatted better that way because it's able to pick out where the deprecation message ends more easily

yen-tt: sure!

nmanu1 (approved):

PR #151, opened 2022-08-16
update search-headless to v2 alpha

oshi97:
Had to update the msw mocks to intercept the new search endpoints.

J=None
TEST=manual

unit tests pass again with no errors
before this there would be an "Invalid API Key" error

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@yext/search-headless-react",
-  "version": "1.4.0",
+  "version": "2.0.0-alpha.151",
   "description": "The official React UI Bindings layer for Search Headless",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -32,7 +32,7 @@
     "generate-notices": "generate-license-file --input package.json --output THIRD-PARTY-NOTICES --overwrite"
   },
   "dependencies": {
-    "@yext/search-headless": "^1.4.0",
+    "@yext/search-headless": "^2.0.0-alpha.130",
```
nmanu1, line 35: can you bump the headless-react version too so we can publish a new alpha?

nmanu1 (approved):
💯 💯

PR #153, opened 2022-08-18
remove deprecated code

juliannzhou:
This PR removes any code deprecated as part of renaming and uses the newest alpha of headless (without ^) where all deprecated code is removed and bump the alpha version of headless to` 2.0.0-alpha.134`.

J=SLAP-2339
TEST=manual

Test: removed any code tagged deprecated and `deprecate.ts` as well as its export in `index.ts`, updated the documentations, and searched 'deprecate', 'answers', as well as all identifiers in Cindy's renaming PRs. Also renamed answers to search for internal function names and changed `answersConfig` to `searchConfig`, since it has been renamed in headless.


package.json
```diff
@@ -32,7 +32,7 @@
     "generate-notices": "generate-license-file --input package.json --output THIRD-PARTY-NOTICES --overwrite"
   },
   "dependencies": {
-    "@yext/search-headless": "^2.0.0-alpha.133",
+    "@yext/search-headless": "2.0.0-alpha.134",
```
nmanu1, line 35: why do we need to pin to this specific alpha version?

juliannzhou: this was addressed per Yen's comment in all-hands.

oshi97 (approved):

PR #155, opened 2022-08-26
Remove the `SearchHeadlessProvider`.

tmeyer2115:
This PR removes the `SearchHeadlessProvider`. With this change, a user will need to invoke `provideHeadless` themselves and pass the result to `SearchHeadlessContext.Provider`. Importantly, this puts the management of Headless instances on the user.

Having the `SearchHeadlessProvider` create an instance of Headless for you is problematic. This Provider needs to be included in a Component's `render` function. So, if the Component is re-rendered, this Provider creates a new (but identical) Headless instance. The previous instances are never garbage-collected.

Rather than adding some sort of cache to `SearchHeadlessProvider`, it was decided to simply remove it. Having users manage their own Headless instances makes our lives a lot easier. It also provides the user with more flexibility.

Note: This is a breaking change! We will need a new major of Headless React. 

TEST=auto

Ran all the unit tests.


src/useSearchActions.ts
```diff
@@ -7,8 +7,8 @@ export type SearchActions = SearchHeadless;
 export function useSearchActions(): SearchActions {
   const searchHeadless = useContext(SearchHeadlessContext);
```
nmanu1, line 8: do we want to update `SearchHeadlessContext` to accept an object of type `SearchActions` instead of `SearchHeadless`? right now, they are the same, which is why this works, but if we extended `SearchActions`, this code could be problematic

PR #156, opened 2022-09-01
Rework the `SearchHeadlessProvider` to be a pass-through.

tmeyer2115:
Per Product's request, the `SearchHeadlessProvider` is now a mere pass-through to the `SearchHeadlessContext.Provider`. Users will need to create and initialize a `Headless` instance themselves. This PR also removed the `Client-SDK` header for Headless React. Product and I determined that it was not useful to track the Headless React version. 

The `acquireSessionId` utility will be moved to either the Component Library or Headless, TBD. The Jest test that was added ensures the `Provider` does not touch anything in the `Headless` instance.

TEST=auto

Ran all unit tests.


tests/SearchHeadlessProvider.test.tsx
```diff
@@ -1,23 +1,19 @@
...
 jest.mock('@yext/search-headless', () => ({
-  provideHeadless: jest.fn(() => ({
-    setSessionTrackingEnabled: jest.fn(),
-    setSessionId: jest.fn()
-  }))
+  provideHeadless: jest.fn(() => { return {}; })
 }));
 
-it('correctly passes through a search config with sandbox endpoints', () => {
+it('does not change anything about the supplied Headless instance', () => {
```
oshi97, line 9: I feel like now just tests that no error is thrown? it doesn't really test that nothing is changed. I was thinking of just changing the name

src/SearchHeadlessProvider.tsx
```diff
@@ -1,33 +1,14 @@
...
-  verticalKey?: string,
-  sessionTrackingEnabled?: boolean
-};
+type Props = { searcher: SearchHeadless };
 
 export function SearchHeadlessProvider(props: PropsWithChildren<Props>): JSX.Element {
-  const { children, verticalKey, sessionTrackingEnabled=true, ...searchConfig } = props;
-  const additionalHttpHeaders = {
-    'Client-SDK': {
-      ANSWERS_HEADLESS_REACT: version
```
nmanu1, line 18: if we don't need to pass this header anymore, we could simplify `provideHeadless` and remove the overloaded signatures in Headless

oshi97: do you remember if we said that users are not allowed to set additionalHttpHeaders? I vaguely remember that being the point of the overloaded signatures but can't remember exactly

nmanu1: yes, that's why we had the overloaded signatures and marked the one that accepted headers as `@internal`. we didn't want to support users adding their own arbitrary headers

tests/SearchHeadlessProvider.test.tsx
tmeyer2115: Because the mocked Headless instance has no methods, an error will be thrown if the `Provider` tries to invoke any method on the user-supplied Headless instance. That's the way I came up with to ensure the `Provider` doesn't much with the Headless instance in any way.

src/SearchHeadlessProvider.tsx
tmeyer2115: I'll root through the Slack convos and see if there were other use cases for the additional headers!

tmeyer2115: Cool! Sounds like we can simplify Headless then!

tests/SearchHeadlessProvider.test.tsx
oshi97: ohh gotcha that makes sense. 
Maybe the test name could be changed to say that we don't call any headless methods within the provider? It's maybe a little literal, but I feel like it's not guaranteed that nothing changes about the headless instance. For instance we could wrap it (maybe to "decorate answers actions?") or add additional properties to it. 

Instead of returning an empty object, we could also return a Proxy that freaks out (throws an error) if it gets modified in any way

oshi97: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Proxy we could add get and set handlers to just throw an Error

oshi97 (approved):

PR #175, opened 2023-06-20
bump version for beta release

saahithjanapati:


package-lock.json
```diff
@@ -51,6 +51,38 @@
         "react": "^16.14 || ^17 || ^18"
       }
     },
+    "../search-headless": {
```
benmcginnis: I think this is the problem, this is a relative path

cea2aj: looks like it's fixed now!

cea2aj (approved):
LGTM!

PR #180, opened 2023-06-22
update package-lock.json

saahithjanapati:


package-lock.json
```diff
@@ -1,12 +1,12 @@
 {
   "name": "@yext/search-headless-react",
-  "version": "2.2.0",
+  "version": "2.2.0-beta.1",
```
cea2aj, line 3: since 2.2.0 is already released, this should be 2.3.0-beta.0

cea2aj (approved):

PR #190, opened 2023-09-25
PagesJS compatibility

cea2aj:
Introduce tsup and improve the package.json file in order to support PagesJS.

This works in both dev and prod mode of pagesJS

J=BACK-2545
TEST=manual

Publish an alpha version and test it with https://github.com/yext/search-ui-react/pull/400 and @yext/pages@1.0.0-rc.3. Add several search components to the page and confirm they work as expected.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,10 +1,11 @@
 {
   "name": "@yext/search-headless-react",
-  "version": "2.4.0-alpha.6",
+  "version": "2.4.0-beta.0",
   "description": "The official React UI Bindings layer for Search Headless",
   "main": "./dist/index.cjs",
   "module": "./dist/index.js",
   "type": "module",
+  "sideEffects": false,
   "license": "BSD-3-Clause",
   "types": "./dist/index.d.ts",
   "keywords": [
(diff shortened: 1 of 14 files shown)


package.json
```diff
@@ -1,10 +1,12 @@
...
-  "version": "2.3.2",
+  "version": "2.4.0-alpha.6",
   "description": "The official React UI Bindings layer for Search Headless",
-  "main": "./lib/esm/index.js",
+  "main": "./dist/index.cjs",
+  "module": "./dist/index.js",
+  "type": "module",
   "license": "BSD-3-Clause",
-  "types": "./lib/esm/index.d.ts",
+  "types": "./dist/index.d.ts",
```
mkilpatrick, line 9: Missing "sideEffects": false,

package.json
```diff
@@ -13,18 +15,20 @@
...
+    "dist",
     "src",
     "THIRD-PARTY-NOTICES"
   ],
   "scripts": {
-    "build": "rm -rf lib/** && npm run build:esm && npm run build:cjs && npm run generate-notices",
-    "build:esm": "tsc -p tsconfig.esm.json",
-    "build:cjs": "tsc -p tsconfig.cjs.json",
+    "build": "rm -rf lib/** && rm -rf dist/** && npm run build:js && npm run generate-notices",
+    "build:js": "tsc -p tsconfig.esm.json && tsc -p tsconfig.cjs.json && tsup src/index.ts --dts --sourcemap --format esm,cjs",
```
mkilpatrick, line 31: I still don't understand the tsc thing. Your argument was that people can import the cjs/esm files specifically at lib/cjs or lib/esm. Doesn't that mean you're effectively building this 4 times now? Twice with tsc and twice with tsup.

cea2aj: yes, that means the it's being built 4 times. If we want to remove two of those, I can remove tsc and copy the build output from dist to lib, or I could add additional package.json files in lib in order to redirect those imports to dist

benmcginnis:
We’d have to do a breaking change right? In case people are using the other
ways

package.json
cea2aj: added

package.json
cea2aj: I removed tsc and copied the files into lib

benmcginnis (approved):

PR #196, opened 2023-12-08
Prepare v2.4.0

nmanu1:
Bump package.json to v2.4.0 and search-headless to v2.5.0.

J=none
TEST=auto

### DIFF
--- a/THIRD-PARTY-NOTICES
+++ b/THIRD-PARTY-NOTICES
@@ -62,7 +62,7 @@ SOFTWARE.
 
 The following NPM package may be included in this product:
 
- - @yext/search-core@2.5.0-beta.3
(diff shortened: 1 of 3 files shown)


benmcginnis (approved):
nice! do you think it will be simple enough to bump the other libraries to support node 18 (and add 20 while we're at it)?

nmanu1:
> nice! do you think it will be simple enough to bump the other libraries to support node 18 (and add 20 while we're at it)?

I'm not anticipating too much difficulty based on the quick check I did. I added it for our unit testing node matrix here and tests pass. But, I think it would be good to make an item to more thoroughly investigate whether we fully support 20 (and to add 18 to the unit testing) in all the libraries in case there are edge cases

